Issue #24, opened 2023-08-07
Error handling vmlinux files


Wmyers559:
Surfactant crashes with an IndexError when attempting to process linux kernel boot images (vmlinux) as it incorrectly assumes that they are Windows PE files and then fails when processing the non-existent 'optional' header.

To reproduce:
 * run `surfactant generate` on a container that has a linux kernel boot image in COFF format
 
The root cause of this is that `filetypeinfo/id_magic.py` assumes any file with the magic bytes `MZ` is a PE file, when `MZ` really denotes a DOS EXE file, which PE is built off of. The COFF linux boot image format starts with the same `MZ`, but notably is lacking the standard DOS stub of 'This program cannot be run in DOS mode` and does not have the windows-specific 'optional' header. A more robust means of checking for PE files could also check for that message, or for the optional header located after the DOS stub and standard COFF headers.

This is system-agnostic.

nightlark:
Thanks for the detailed bug report! This week is rather busy, but I'll try to get a fix in by the end of next week.

There should also really be a set of test cases malformed files to make sure the different libraries used to handle various file formats don't crash Surfactant.

Wmyers559:
I had spent a little bit of time looking at this, and have a WIP fork where I'm doing some more robust checking of the kernel header itself.

PR #28, opened 2023-08-14
Fix crashing on vmlinux files

Wmyers559:
Allow for processing non-PE files that use the same DOS format, like linux vmlinux/boot images. The full linux boot image header description can be found at https://www.kernel.org/doc/Documentation/x86/boot.txt

This fixes issue #24; feel free to add more robust checking as needed/desired. Given the kernel documentation, it should be pretty easy to write out a plugin to get the various properties for a SBOM. Perhaps a new feature request :smile: 

If desired, I can see about finding a test file for the vmlinux portion, but this does manage to fix the crash without seeming to cause other issues.


surfactant/filetypeid/id_magic.py
```diff
@@ -28,7 +28,22 @@ def identify_file_type(filepath: str) -> Optional[str]:
             if magic_bytes[:4] == b"\x7fELF":
                 return "ELF"
             if magic_bytes[:2] == b"MZ":
+                # Check to see if this is actually a PE file or a DOS file
+                coff_addr = (
+                    int.from_bytes(magic_bytes[0x3C:0x40], byteorder="little", signed=False)
+                    & 0xFFFF
+                )
+                if magic_bytes[coff_addr : coff_addr + 2] != b"PE":
+                    return "DOS"
```
nightlark, line 37:
```suggestion
                if magic_bytes[coff_addr : coff_addr + 4] != b"PE\x00\x00":
                    return "DOS"
```

It looks like the expected magic for a PE file is `PE\x00\x00` -- it would be interesting to know if Windows will accept a PE file without those two trailing null-bytes as part of the magic bytes or not.

The other potential corner case would be if the coff_addr it gets from the file points to a location beyond the initial 265 byte read; it seems like the expected coff_addr is basically a fixed value, so this would be another unlikely case that would only appear in an intentionally odd or malformed PE file.

surfactant/filetypeid/id_magic.py
```diff
@@ -28,7 +28,22 @@ def identify_file_type(filepath: str) -> Optional[str]:
...
+
+                # Check for the linux kernel header at 0x202 (requires a second read)
+                magic_bytes = f.read(265)
+                if magic_bytes[249:253] == b"HdrS":
```
nightlark, line 41:
```suggestion
                magic_bytes += f.read(265)
                if magic_bytes[0x202:0x206] == b"HdrS":
```

Supporting file checks that require accessing data past the initial read is something I've been meaning to figure out -- I think for the subsequent read appending the new data to the previous bytes read could make it easier to follow what offset in the file is being checked. Something similar could be done for ensuring enough data is read when the offset to check if specified by another field in the header (e.g. `coff_addr`).

nightlark (approved):
Thanks for the fixes! I'll wait a bit before merging to give you an opportunity to comment on the suggestions I left -- they don't need to be made in this PR, but tie into possible refactoring of the magic bytes recognition code that I've been contemplating.

surfactant/filetypeid/id_magic.py
Wmyers559: Definitely, that's something that I was thinking about, but wasn't sure how you all would like to best handle that. I think that this looks good -- I just used another 265 byte read as that was the same size as the initial read.

Wmyers559:
I added in the changes; they were things that I had been thinking about anyway, and thanks for catching the `PE` vs `PE\x00\x00` distinction.

surfactant/filetypeid/id_magic.py
```diff
@@ -28,7 +28,36 @@ def identify_file_type(filepath: str) -> Optional[str]:
...
+                coff_addr = (
+                    int.from_bytes(magic_bytes[0x3C:0x40], byteorder="little", signed=False)
+                    & 0xFFFF
+                )
+
+                # Check to see if the coff_addr is still within the initial
+                # read; if not, we need at least another one, maybe more.
+                # Unlikely, but possible if the PE/file is malformed.
+                if coff_addr > len(magic_bytes):
+                    magic_bytes += f.read(265)
```
nightlark, line 41:
```suggestion
                if coff_addr+4 > len(magic_bytes):
                    magic_bytes += f.read(coff_addr+4-len(magic_bytes))
```

Since this is likely to already be an odd/malformed PE file, we might as well attempt a read from the file that gets all of the magic bytes for the PE header.

surfactant/filetypeid/id_magic.py
```diff
@@ -28,7 +28,36 @@ def identify_file_type(filepath: str) -> Optional[str]:
...
+
+                elif coff_addr > len(magic_bytes) * 2:
```
nightlark, line 43:
```suggestion
                if coff_addr+4 > len(magic_bytes):
```

If the bytes read at this point still aren't enough to reach `coff_addr`, then the file is definitely a malformed PE file or another file type with something strange going on.

nightlark:
Thanks! Just two minor changes to the PE/COFF address read and malformed PE check. If you want to do a rebase and manual squash that includes `(#28)` at the end of the commit title that would be great -- otherwise there aren't any conflicts so GitHub's squash and merge should work.

surfactant/filetypeid/id_magic.py
Wmyers559: I've added this in, good catch. This, for example, could be a pure DOS file without the extra COFF header and some random or strange value in that field.

nightlark:
Awesome, merged!

> Given the kernel documentation, it should be pretty easy to write out a plugin to get the various properties for a SBOM. Perhaps a new feature request 😄

Could you create an issue or a GitHub discussion thread to track that idea?

> If desired, I can see about finding a test file for the vmlinux portion, but this does manage to fix the crash without seeming to cause other issues.

If there's a relatively small vmlinux file that could be used for testing that would be great to have for a future plugin that extracts information from linux kernel images.

PR #29, opened 2023-08-14
Cyt 444 more unit tests

KendallHarterAtWork:
New tests:
PE relationships
.net relationships
magic number identification
a.out file identification


tests/symlink/test_resolve_links.py
```diff
@@ -33,27 +33,28 @@ def create_symlinks(temp_dir):
 
 
 def test_symlinks():
```
nightlark, line 35: Does the pytest skipif decorator work? It might require importing `pytest`.

```suggestion
@pytest.mark.skipif(os.name != 'posix',
                    reason="requires posix os")
def test_symlinks():
```

nightlark (approved):

PR #39, opened 2023-08-25
Add wheel package to build-system requirements

### LINKED ISSUE #25: Pypi installer fails if wheel is not installed
If following the instructions in the original readme, the build will fail in a virtual environment if wheel is not installed which then requires the user to manually install wheel before finishing the installation.

nightlark:
It seems like some systems (likely with older versions of Python or pip) may require `wheel` to be listed as a build-system requirement for installing Surfactant from source.

@RazerNinjas -- check if this works on the system you were getting the failures on.

Resolves #25.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["setuptools", "setuptools-scm"]
+requires = ["setuptools", "setuptools-scm", "wheel"]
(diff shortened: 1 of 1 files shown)


nightlark:
`main` branch worked when tested -- closing for now, until it happens again and more details on the system/environment needed to reproduce the error can be gathered.

PR #40, opened 2023-08-28
Cyt 383 support running external tools syft

CCBromia:
Merging the Syft plugin into the main branch.


nightlark:
I'll do another pass through this soon and try to get it merged. What I'm currently thinking is #83 will get merged first which makes some changes to the core directory traversal loop and information passed to the extract_file_info hook to enable additional analysis of files that a plugin might extract from another one -- there could be some ways to use that instead of creating a new hookspec that is fairly similar.

surfactantplugin-syft/surfactantplugin_syft.py
```diff
@@ -0,0 +1,98 @@
...
+import time
+from typing import List, Optional
+
+import surfactant.plugin
+from surfactant.plugin.manager import get_plugin_manager
+from surfactant.sbomtypes import SBOM, Relationship, Software
+
+
+@surfactant.plugin.hookimpl
+def extract_child_info(
```
nightlark, line 12:
```suggestion
def extract_file_info(
```
Switch to using the existing hook.

surfactantplugin-syft/surfactantplugin_syft.py
```diff
@@ -0,0 +1,98 @@
...
+            )
+            for file in i["metadata"]["files"]:
+                if file["path"] not in sw_entry.supplementaryFiles:
+                    sw_entry.supplementaryFiles.append(file["path"])
+            for file in i["locations"]:
+                if file["path"] not in sw_entry.supplementaryFiles:
+                    sw_entry.supplementaryFiles.append(file["path"])
+            sw_list.append(sw_entry)
+        gather_relationship_data(software, data, sw_list)
+        return sw_list
```
nightlark, line 54:
```suggestion
            children.append(sw_entry)
        gather_relationship_data(software, data, children)
```

surfactantplugin-syft/surfactantplugin_syft.py
```diff
@@ -0,0 +1,98 @@
...
+            "anchore_syft " + filename + " -o json --scope all-layers", shell=True
+        )
+        data = json.loads(data.decode())
+        sw_list = []
+        for i in data["artifacts"]:
+            sw_entry = Software(
+                sha1=None,
+                # Syft does not provide a SHA256 for each artifact.
+                # This uses their unique IDs in place since surfactant is dependent on this hash.
+                sha256=i["id"],
```
nightlark, line 28: Files from within a tar file are a bit of an odd case. They should be able to get their own software entry rather than be a component; but Syft doesn't provide a hash for the files in it, which is awkward. For this let's add a TODO for me to discuss with Lucas. Or maybe better, open a GitHub issue rather than adding it as a comment.

```suggestion
                # Syft does not provide a SHA256 for each artifact.
                # This uses their unique IDs in place since surfactant is dependent on this hash.
                # TODO (Ryan) discuss how to handle software entry without a hash with Lucas in the CyTRICS SBOM schema
                sha256=i["id"],
```

nightlark (commented):
Let's merge the functionality for adding nested children software entries into the existing `extract_file_info` hook instead of creating a new one -- I think it will be less confusing for plugin authors than having two fairly similar hooks to choose from. It also enables a single plugin to add both metadata and children software entries, rather than having to be split across two different hooks(/running twice).

nightlark:
@CCBromia if the changes look okay with you, I can take care of applying + testing them.

nightlark (approved):

PR #42, opened 2023-09-06
Use extractPaths as installPrefix when not given

### LINKED ISSUE #32: Generate "Uses" Relationships without installPrefix
Note: I have framed this as a feature request as I'm unsure if the current behavior is a bug or expected.

**Is your feature request related to a problem? Please describe.**
When I run Surfactant using a config file that only contains an `extractPaths` entry then no "Uses" relationships are generated for imports. To get the "Uses" relationships to be generated I must include an `installPrefix`. This requirement is not currently documented, as far as I can see, so I'm not sure if this is expected.

As I initially read the documentation, I expected `installPrefix` to help in two situations. One, be a way of overriding the generated SBOM's `installPath` to better reflect an actual installation of the software. Two, it would also give the user an opportunity to remove any local information that may be revealed in the extract path. I didn't expect that I had to specify `installPrefix` in order to generate relationships.

**Describe the solution you'd like**
My ideal solution is that the `extractPaths` are used as the `installPath` if an `installPrefix` is not explicitely stated. This would ensure that "Uses" relationships are generated by default. I would include the `extractPaths` in the generated `installPath` of the SBOM unless overriden by the `installPrefix`.

This solution would break existing behavior in two ways:
1. Generate relationships when they used to not.
2. Generate an `installPath` when it used to be an empty list.

Change 1 can be addressed by the user using `--skip_relationships`. Change 2 cannot be easily addressed and I believe that is still acceptable. The current implementation generates no `installPath` information if an `installPrefix` is not provided.
That makes it so that file system hierarchical information is not collected (i.e. the generated SBOM appears to exist in a flat file system). I don't believe that this is especially useful by default and users would rather collect that information without additional configuration.

**Describe alternatives you've considered**
One alternative solution would be to use the extract path solely for the sake of calculating relationships but to not save it as the `installPath`. I believe this would be a more complicated solution as current relationship generating methods appear to rely on the `installPath` field in the software object.

Another solution is to simply update the documentation to better reflect that an `installPrefix` must be provided to ensure that "Uses" relationships are captured.

**Additional context**
If you run Surfactant by passing an extract path directly through the CLI (instead of using a config file), then the extract path is automatically used as the `installPrefix` and relationships are generated. This behavior matches my proposed solution.

I have found that my documentation of existing behavior is consistent on Windows and Linux for PE and ELF binaries, respectively.

The existing documentation does seem to imply the existing behavior is expected. Example 1 in the README demonstrates a bare config file with no `installPrefix` and the generated SBOM has no relationships. Example 2 then shows a config file with an `installPrefix` and the relationships are in fact generated. I would suggest that this behavior be more explicitely stated if it is expected though.

douglasdennis:
This PR makes the following changes:

- If a configuration file does not provide an `installPrefix` then the `extractPaths` will be used to generate the `installPath` of each enumerated binary. This makes it so that "Uses" relationships will be generated by default when no `installPath` is given.
- A new flag, **--skip_install_path** has been added. This flag will turn off the new default behavior so that if an `installPrefix` is not given then no `installPath` will be generated.
- The README has been updated to document the new flag and the new behavior.
- Tests for the generate command have been added that demonstrate the new behavior.

A couple notes:
- I wasn't sure how tests were expected to be added in this situation, so I just took a shot at it. I'm open to refactoring them or removing them as needed.
- The tests invoke the generate command programmatically, but they do not use the `click.testing` module. This is because click's testing module shuts off writing to the filesystem so the sbom could not be generated. A future addition to allow for output to be redirected to stdout could help with this.
- I tried testing with situations that used symlinks and what I could do seemed to work for me. However, I admit that I am not confident that I tested this thoroughly.

Closes #32

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,13 +83,13 @@ The resulting SBOM would be structured like this:
         {
           "UUID": "abc1",
           "fileName": ["helics_binary"],
-          "installPath": "/home/samples/helics/bin",
+          "installPath": ["/home/samples/helics/bin/helics_binary"],
           "containerPath": null
         },
         {
           "UUID": "abc2",
           "fileName": ["lib1.so"],
-          "installPath": "/home/samples/helics/bin",
+          "installPath": ["/home/samples/helics/bin/lib1.so"],
           "containerPath": null
         }
 
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ A configuration file contains the information about the sample to gather informa
 
 **extractPaths**: (required) the absolute path or relative path from location of current working directory that `surfactant` is being run from to the sample folders, cannot be a file (Note that even on Windows, Unix style `/` directory separators should be used in paths)\
 **archive**: (optional) the full path, including file name, of the zip, exe installer, or other archive file that the folders in **extractPaths** were extracted from. This is used to collect metadata about the overall sample and will be added as a "Contains" relationship to all software entries found in the various **extractPaths**\
-**installPrefix**: (optional) where the files in **extractPaths** would be if installed correctly on an actual system i.e. "C:/", "C:/Program Files/", etc (Note that even on Windows, Unix style `/` directory separators should be used in the path)
+**installPrefix**: (optional) where the files in **extractPaths** would be if installed correctly on an actual system i.e. "C:/", "C:/Program Files/", etc (Note that even on Windows, Unix style `/` directory separators should be used in the path). If not given then the **extractPaths** will be used as the install paths
 
 #### Example configuration file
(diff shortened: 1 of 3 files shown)


douglasdennis:
This is failing tests due to #44 . Working on addressing that as well.

douglasdennis:
Rebased this branch and all of the tests pass. This PR is ready for a review.

nightlark (approved):
Awesome, thanks for making this change and adding some tests -- just a minor documentation path tweak and it looks ready to merge.

nightlark:
Sorry for the delay in getting this merged -- I missed the notification about the latest commit.

PR #49, opened 2023-10-03
Add hash collision detection

KendallHarterAtWork:


surfactant/cmd/generate.py
```diff
@@ -102,6 +102,31 @@ def print_output_formats(ctx, _, value):
...
+        collision = True
+    elif soft1.md5 == soft2.md5 and (soft1.sha256 != soft2.sha256 or soft1.sha1 != soft2.sha1):
+        collision = True
+    elif (
+        soft1.sha256 == soft2.sha256
+        and soft1.sha1 == soft2.sha1
+        and soft1.md5 == soft2.md5
+        and soft1.size != soft2.size
+    ):
+        collision = True
```
nightlark, line 123:
```
if s1.sha256 == s2.sha256 or s1.sha1 == s2.sha1 or s1.md5 == s2.md5:
    if s1.sha256 and s2.sha256 and s1.sha256 != s2.sha256:
        collision = True
    ...
    elif s1.size != s2.size
        collision =True
```
From what I recall of the CyTRICS schema, it's possible for hashes to be left out (e.g. user supplied an initial likely manual SBOM that generate is adding onto) so some checks to make sure both entries aren't a `None` or empty value before checking for them not being equal would be good.

Maybe an initial outer if statement could confirm that one or more of the hashes matches, before the nested if statements determine if there are any mistmatches?

nightlark (approved):

PR #54, opened 2023-10-09
Added support for ImplMap entries and relationships for .NET binaries

Czatar:
Output for the "dotnetImplMap" key in the metadata shows the name of the DLL paired with a list of functions used from the DLL. Determining relationships works the same way as "dotnetAssemblyRef" entries.

### DIFF
--- a/surfactant/infoextractors/pe_file.py
+++ b/surfactant/infoextractors/pe_file.py
@@ -24,7 +24,9 @@ def supports_file(filetype) -> bool:
 
 
 @surfactant.plugin.hookimpl
-def extract_file_info(sbom: SBOM, software: Software, filename: str, filetype: str) -> object:
+def extract_file_info(
+    sbom: SBOM, software: Software, filename: str, filetype: str
+) -> object:
     if not supports_file(filetype):
         return None
     return extract_pe_info(filename)
@@ -100,10 +102,14 @@ def extract_pe_info(filename):
             "peSubsystemVersion"
         ] = f"{pe.OPTIONAL_HEADER.MajorSubsystemVersion}.{pe.OPTIONAL_HEADER.MinorSubsystemVersion}"
         if pe.OPTIONAL_HEADER.Subsystem in pe_subsystem_types:
-            file_details["peSubsystem"] = pe_subsystem_types[pe.OPTIONAL_HEADER.Subsystem]
+            file_details["peSubsystem"] = pe_subsystem_types[
+                pe.OPTIONAL_HEADER.Subsystem
+            ]
         else:
             file_details["peSubsystem"] = pe.OPTIONAL_HEADER.Subsystem
-            print("[WARNING] Unknown Windows Subsystem type encountered in PE file header")
+            print(
+                "[WARNING] Unknown Windows Subsystem type encountered in PE file header"
+            )
         file_details[
             "peLinkerVersion"
         ] = f"{pe.OPTIONAL_HEADER.MajorLinkerVersion}.{pe.OPTIONAL_HEADER.MinorLinkerVersion}"
@@ -131,18 +137,26 @@ def extract_pe_info(filename):
     if opt_hdr := getattr(pe, "OPTIONAL_HEADER", None):
         if opt_hdr_data_dir := getattr(opt_hdr, "DATA_DIRECTORY", None):
             # COM Descriptor, used to identify CLR/.NET binaries
-            com_desc_dir_num = dnfile.DIRECTORY_ENTRY["IMAGE_DIRECTORY_ENTRY_COM_DESCRIPTOR"]
+            com_desc_dir_num = dnfile.DIRECTORY_ENTRY[
+                "IMAGE_DIRECTORY_ENTRY_COM_DESCRIPTOR"
+            ]
             com_desc_dir = opt_hdr_data_dir[com_desc_dir_num]
-            file_details["peIsClr"] = (com_desc_dir.VirtualAddress > 0) and (com_desc_dir.Size > 0)
+            file_details["peIsClr"] = (com_desc_dir.VirtualAddress > 0) and (
+                com_desc_dir.Size > 0
+            )
 
     if pe_fi := getattr(pe, "FileInfo", None):
         if len(pe_fi) > 0:
             file_details["FileInfo"] = {}
             for fi_entry in pe_fi[0]:
-                if fi_entry.name == "StringFileInfo" and hasattr(fi_entry, "StringTable"):
+                if fi_entry.name == "StringFileInfo" and hasattr(
+                    fi_entry, "StringTable"
+                ):
                     for st in fi_entry.StringTable:
                         for st_entry in st.entries.items():
-                            file_details["FileInfo"][st_entry[0].decode()] = st_entry[1].decode()
+                            file_details["FileInfo"][st_entry[0].decode()] = st_entry[
+                                1
+                            ].decode()
 
     # If this is a .NET assembly, extract information about it from the headers
     if dnet := getattr(pe, "net", None):
@@ -170,7 +184,7 @@ def extract_pe_info(filename):
             if implmap_info := getattr(dnet_mdtables, "ImplMap", None):
                 imp_modules = []
                 for im_info in implmap_info:
-                    get_implmap_info(im_info, imp_modules)
+                    insert_implmap_info(im_info, imp_modules)
                 file_details["dotnetImplMap"] = imp_modules
 
     # TODO for a custom intermediate SBOM format, the information read from the manifest and app config files
@@ -238,7 +252,7 @@ def get_assemblyref_info(asmref_info):
     return asmref
 
 
-def get_implmap_info(im_info, imp_modules):
+def insert_implmap_info(im_info, imp_modules):
     dllName = im_info.ImportScope.row.Name
     methodName = im_info.ImportName
     if dllName:
@@ -262,7 +276,9 @@ def get_xmlns_and_tag(uri):
 # return any info that could be useful for establishing "Uses" relationships later
 def get_windows_manifest_info(filename):
     binary_filepath = pathlib.Path(filename)
-    manifest_filepath = binary_filepath.with_suffix(binary_filepath.suffix + ".manifest")
+    manifest_filepath = binary_filepath.with_suffix(
+        binary_filepath.suffix + ".manifest"
+    )
     if manifest_filepath.exists():
         print("Found application manifest file for " + filename)
         et = defusedxml.ElementTree.parse(manifest_filepath)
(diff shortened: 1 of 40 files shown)


surfactant/infoextractors/pe_file.py
```diff
@@ -233,6 +238,17 @@ def get_assemblyref_info(asmref_info):
     return asmref
 
 
+def get_implmap_info(im_info, imp_modules):
```
nightlark, line 241:
```suggestion
def get_implmap_info(im_info, imp_modules):
```

This is doing a bit more than just getting the implmap -- maybe call it `add_implmap_info` or `insert_implmap_info` instead to capture that it is also modifying part of the SBOM.

surfactant/relationships/dotnet_relationship.py
```diff
@@ -75,8 +75,16 @@ def establish_relationships(
...
     #    - the location of the calling assembly may be used as a hint for where to find the referenced assembly
+    def chained_lists(*it):
+        for iterab in it:
+            yield from iterab
+
     if "dotnetAssemblyRef" in metadata:
-        for asmRef in metadata["dotnetAssemblyRef"]:
+        combinedRefs = [metadata["dotnetAssemblyRef"]]
+        if "dotnetImplMap" in metadata:
+            combinedRefs.append(metadata["dotnetImplMap"])
```
nightlark, line 85: What were the URLs/doc links that you referenced for info on how .NET determines where to look for the DLLs listed in ImplMap?

Czatar: I was mistaken about unmanaged code loading being the same as loading assemblies. I got my sources mixed up, but I found the one that relates to ImplMap (PInvoked) libraries and included it. From what it looks like, if it's not the assembly/application directory, it's the system32 folder (default unless someone overrides the loading function).

surfactant/relationships/dotnet_relationship.py
```diff
@@ -65,6 +65,34 @@ def establish_relationships(
...
+    # https://learn.microsoft.com/en-us/dotnet/core/dependency-loading/loading-unmanaged
+    # 1. Check the active AssemblyLoadContext cache
+    # 2. Calling the import resolver set by the setDllImportResolver function
+    #    - a. Example using SetDllImportResolver: https://learn.microsoft.com/en-us/dotnet/standard/native-interop/native-library-loading
+    #    - b. Checks PInvoke's or Assembly's DefaultDllImportSearchPathsAttribute, then the assembly's directory, then LoadLibraryEx with LOAD_WITH_ALTERED_SEARCH_PATH flag
+    #       - This attribute has no effect on non-Windows platforms / Mono runtime
+    #       - https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.defaultdllimportsearchpathsattribute?view=net-7.0
+    #       - i. This has a "Paths" property which is a bitwise combination of paths specified in ii:
+    #       - ii. https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.dllimportsearchpath?view=net-7.0
+    # 3. The active AssemblyLoadContext calls its LoadUnmanagedDll function (Default behavior is the same as AssemblyRef probing?)
```
nightlark, line 77: Based on https://learn.microsoft.com/en-us/dotnet/core/dependency-loading/default-probing#satellite-resource-assembly-probing it seems like it is -- though it's hard to tell if that satellite assembly probing section applies to loading unmanaged libraries as well. It's possible that the default probing logic is just https://learn.microsoft.com/en-us/dotnet/core/dependency-loading/default-probing#unmanaged-native-library-probing and none of the AssemblyRef probing stuff applies.

surfactant/relationships/dotnet_relationship.py
```diff
@@ -65,6 +65,34 @@ def establish_relationships(
...
+            refName = None
+            refCulture = None
+            if "Name" in asmRef:
+                refName = asmRef["Name"]
+
+            # Aside from .deps.json, the assembly's install directory is searched
+            probedirs = get_dotnet_probedirs(software, refCulture, refName, dnProbingPaths)
+            for e in find_installed_software(sbom, probedirs, refName + ".dll"):
+                dependency_uuid = e.UUID
+                relationships.append(Relationship(dependent_uuid, dependency_uuid, "Uses"))
```
nightlark, line 94: In addition to probing directories similar to .NET (or instead of depending on if we can determine whether or not this applies to loading unmanaged libraries), it looks like we need to have code that looks for relationships according to https://learn.microsoft.com/en-us/dotnet/core/dependency-loading/default-probing#unmanaged-native-library-probing and searches for names/paths according to https://learn.microsoft.com/en-us/dotnet/standard/native-interop/native-library-loading when a relative path is given.

surfactant/relationships/dotnet_relationship.py
```diff
@@ -65,6 +66,58 @@ def establish_relationships(
...
+    # 4. Run default unmanaged library probing logic by parsing *.deps.json probing properties
+    #    - a. If the json file isn't present, assume the calling assembly's directory contains the library
+    #    - b. https://learn.microsoft.com/en-us/dotnet/core/dependency-loading/default-probing#unmanaged-native-library-probing
+    if "dotnetImplMap" in metadata:
+        for asmRef in metadata["dotnetImplMap"]:
+            if "Name" not in asmRef:
+                continue
+            refName = asmRef["Name"]
+            if is_absolute_path(refName):
+                relationships.extend(get_windows_pe_dependencies(sbom, software, [refName]))
```
nightlark, line 90: I don't think the `get_windows_pe_dependencies` function does what we want when a dependency is specified using absolute paths -- the function itself constructs a list of probe directories relative to the software, which when the given path is absolute we don't care about searching, and then it calls the `find_installed_software` function internally to find a match, which appends the file name for an import (which in this case would be an absolute path) to each probe directory given in turn.

For an example of a program in `C:/PrgFiles/Program.exe` an absolute import of `C:/SomeFolder/Dep.dll` looks like it would result in the function trying to search for something like `C:/PrgFiles/C:/Somefolder/Dep.dll`.

For matching an import listed as an absolute dependency, I think we can have a loop that just looks for a Software entry with an install path that just matches the absolute import name, without any of the probe directory things in it.

```python
ref_abspath = pathlib.PureWindowsPath(refName)
# iterate through all sbom entries
for e in sbom.software:
    # Skip if no install path (e.g. installer/temporary file)
    if e.installPath is None:
        continue
    if isinstance(e.installPath, Iterable):
        for ifile in e.installPath:
            # PureWindowsPath is case-insensitive for file/directory names
            if ref_abspath == pathlib.PureWindowsPath(ifile):
                # matching probe directory and filename, add relationship to list
                relationships.extend(Relationship(dependent_uuid, e.uuid, "Uses"))
```

surfactant/relationships/dotnet_relationship.py
```diff
@@ -65,6 +66,58 @@ def establish_relationships(
...
+                combinations.append(f"{refName}.dll")
+            combinations.extend(
+                [
+                    f"{refName}.so",
+                    f"{refName}.dylib",
+                    f"lib{refName}.so",
+                    f"lib{refName}.dylib",
+                    f"lib{refName}",
+                ]
+            )
```
nightlark, line 107: I'm not familiar with the actual behavior for how .NET will probe for files here, and the online doc page doesn't seem to touch on this -- if refName is `abc.dll`, then on Linux will it search for a file named `abc.dll.so`, or does .NET try to be smart and remove the `.dll` so before appending `.so`?

surfactant/relationships/dotnet_relationship.py
```diff
@@ -65,6 +66,58 @@ def establish_relationships(
...
+            if ".so." in refName or refName.endswith(".so"):
+                combinations.extend(
+                    [
+                        f"{refName}.so.6",
+                        f"lib{refName}.so.6",
+                        f"{refName}.so.6.so",
+                        f"lib{refName}.so.6.so",
+                    ]
+                )
```
nightlark, line 116: Can you create a GitHub Issue about adding support for "fuzzy" dependency name matching using regular expressions? The `6` is actually a version related number for the shared library, so these should actually be matching against any number with a regex like `f"{refname}.so.[0-9]?"`. That's a fairly large overhaul of the functions we currently have for finding dependencies, so shouldn't be done right now.

These numbers in a shared library file name also tend to be associated with symlinks, so there is likely to be a plain (e.g.) `{refname}.so` file that will be matched because it symlinks to (e.g.) `{refname}.so.6`.

surfactant/relationships/dotnet_relationship.py
nightlark: Decided on re-reading the page that satellite-resource-assembly-probing does not apply.

surfactant/relationships/dotnet_relationship.py
nightlark: Issue #80 - backlog for something to check eventually

surfactant/relationships/dotnet_relationship.py
nightlark: #79 - backlog issue for adding fuzzy/regex dependency name matching support.

surfactant/relationships/dotnet_relationship.py
```diff
@@ -86,13 +88,29 @@ def establish_relationships(
...
+                        continue
+                    if isinstance(e.installPath, Iterable):
+                        for ifile in e.installPath:
+                            if ref_abspath == pathlib.PureWindowsPath(ifile):
+                                relationships.extend(
+                                    Relationship(dependent_uuid, e.uuid, "Uses")
+                                )
+                relationships.extend(
+                    get_windows_pe_dependencies(sbom, software, [refName])
+                )
```
nightlark: I don't think this line is needed -- haven't confirmed, but in that function it is appending refName (which in this case is an absolute path such as `C:/Somefolder/Dep.dll`) with the path software is located in this function will just be searching for garbage paths like `C:/PrgFiles/C:/Somefolder/Dep.dll` that are actually invalid path names on Windows (`:` is not allowed in folder names) and can't ever exist.

surfactant/relationships/dotnet_relationship.py
```diff
@@ -65,6 +66,62 @@ def establish_relationships(
...
+            if is_absolute_path(refName):
+                ref_abspath = pathlib.PureWindowsPath(refName)
+                for e in sbom.software:
+                    if e.installPath is None:
+                        continue
+                    if isinstance(e.installPath, Iterable):
+                        for ifile in e.installPath:
+                            if ref_abspath == pathlib.PureWindowsPath(ifile):
+                                relationships.extend(Relationship(dependent_uuid, e.uuid, "Uses"))
+                relationships.extend(get_windows_pe_dependencies(sbom, software, [refName]))
```
nightlark, line 100: I don't think the line calling `get_windows_pe_dependencies` should be present - haven't confirmed, but in that function it is appending refName (which in this case is an absolute path such as C:/Somefolder/Dep.dll) with the path software is located in, so it will just be searching for garbage paths like C:/PrgFiles/C:/Somefolder/Dep.dll that are actually invalid path names on Windows (: is not allowed in folder names) so the files it is looking for should never exist.

nightlark (approved):

PR #55, opened 2023-10-10
Added hook for reading CyTRICS format input SBOMs

mws180000:
Implemented hook for reading input SBOMs that are in the (default) CyTRICS format.

TODO: Add additional hooks for input SBOMs in SPDX and CycloneDX formats.

### DIFF
--- a/surfactant/cmd/generate.py
+++ b/surfactant/cmd/generate.py
@@ -6,13 +6,12 @@ import json
 import os
 import pathlib
 import re
-import sys
 from typing import Dict, List, Optional, Tuple, Union
 
 import click
 from loguru import logger
 
-from surfactant.plugin.manager import get_plugin_manager
+from surfactant.plugin.manager import find_io_plugin, get_plugin_manager
 from surfactant.relationships import parse_relationships
 from surfactant.sbomtypes import SBOM, Software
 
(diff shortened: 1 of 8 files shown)


surfactant/cmd/generate.py
```diff
@@ -186,7 +231,7 @@ def sbom(
     if not input_sbom:
         new_sbom = SBOM()
     else:
-        new_sbom = SBOM.from_json(input_sbom.read())
+        new_sbom = input_reader.read_sbom(input_sbom)
```
nightlark, line 234: Can you make this same change for the merge command in https://github.com/LLNL/Surfactant/blob/CYT-470-add-hook-for-reading-sbom-files/surfactant/cmd/merge.py#L33?

mws180000: I have finished making those change to the merge command. For now, should we assume that all the input SBOMs being merged have the same format?

nightlark: Yes, for now let’s assume they all have the same format, and update docs or argument descriptions to mention that.

surfactant/cmd/merge.py
```diff
@@ -20,17 +21,39 @@
...
+                if plugin.short_name().lower() == input_format.lower() and hasattr(
+                    plugin, "read_sbom"
+                ):
+                    input_reader = plugin
+                    break
+            except AttributeError:
+                pass
+
+    if input_reader is None:
+        logger.error(f'No output format "{input_format}" found')
```
nightlark, line 51:
```suggestion
        logger.error(f'No input reader plugin for format "{input_format}" found')
```

nightlark: Note: This pattern of searching for input/output readers seems like something that is going to come up in several places -- may be useful to consider moving most of the code into a plugin manager helper function that searches for a plugin with the given plugin name and implementing a given hooked function name ("read_sbom").

surfactant/cmd/generate.py
mws180000: I have updated the relevant READMEs

surfactant/cmd/merge.py
mws180000: I have moved the code for finding input readers and output writers into a separate function in manager.py

nightlark (approved):

PR #56, opened 2023-10-17
create-config feature added

theStache:
A basic configuration file can be created using the 'create-config' command. This will take a path as a command line argument and will save a file with the default name of the end directory passed to it as a configuration json file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,7 +13,6 @@ __pycache__/
 
 # C extensions
 *.so
-cytrics_venv
 
 # Distribution / packaging
 .Python
@@ -136,6 +135,7 @@ venv/
 ENV/
 env.bak/
 venv.bak/
+cytrics_venv
 
 # Spyder project settings
 .spyderproject
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@ __pycache__/
 
 # C extensions
 *.so
+cytrics_venv
 
 # Distribution / packaging
 .Python
--- a/surfactant/cmd/createconfig.py
+++ b/surfactant/cmd/createconfig.py
@@ -24,7 +24,7 @@ def create_config(directory, output):
         # If there are no files, add the immediate subdirectories
         for item in starting_directory.iterdir():
             if item.is_dir():
-                extract_paths.append(str(item).replace("\\", "/"))
+                extract_paths.append(item.as_posix())
 
     config_dict = {"extractPaths": extract_paths, "installPrefix": "/"}
     config_out = [config_dict]
--- a/surfactant/cmd/createconfig.py
+++ b/surfactant/cmd/createconfig.py
@@ -1,4 +1,5 @@
 import json
+from loguru import logger
 from pathlib import Path
 
 import click
--- a/surfactant/cmd/createconfig.py
+++ b/surfactant/cmd/createconfig.py
@@ -21,24 +21,10 @@ from loguru import logger
 )
 def create_config(directory, output, install_prefix):
     """Create surfactant input configuration file based on input directory."""
-    starting_directory = Path(directory)
-    extract_paths = []
-
-    # Check if there are any files in the directory
-    if any(item.is_file() for item in starting_directory.iterdir()):
-        extract_paths.append(starting_directory.as_posix())
-    else:
-        # If there are no files, add the immediate subdirectories
-        for item in starting_directory.iterdir():
-            if item.is_dir():
-                extract_paths.append(item.as_posix())
-
+    extract_paths = [Path(directory).as_posix()]
     config_dict = {"extractPaths": extract_paths, "installPrefix": install_prefix}
     config_out = [config_dict]
-
-    output_file_name = output or starting_directory.stem + ".json"
-
+    output_file_name = output or Path(directory).stem + ".json"
     with open(output_file_name, "w") as json_file:
         json.dump(config_out, json_file, indent=4)
-
     logger.info(f"Data written to {output_file_name}")
--- a/surfactant/cmd/createconfig.py
+++ b/surfactant/cmd/createconfig.py
@@ -1,8 +1,8 @@
 import json
-from loguru import logger
 from pathlib import Path
 
 import click
+from loguru import logger
 
 
 @click.command()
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -13,6 +13,7 @@ __pycache__/
 
 # C extensions
 *.so
+cytrics_venv
```
nightlark, line 16: nit - I'd probably add this to the `Environments` section that begins around line 131 of this file.

surfactant/cmd/createconfig.py
```diff
@@ -0,0 +1,37 @@
...
+    else:
+        # If there are no files, add the immediate subdirectories
+        for item in starting_directory.iterdir():
+            if item.is_dir():
+                extract_paths.append(str(item).replace("\\", "/"))
+
+    config_dict = {"extractPaths": extract_paths, "installPrefix": "/"}
+    config_out = [config_dict]
+
+    output_file_name = output or starting_directory.stem + ".json"
```
nightlark, line 32:
```suggestion
    output_file_name = output or starting_directory.with_suffix(".json")
```

Should be functionally equivalent, but with some extra error checking for invalid names.

surfactant/cmd/createconfig.py
```diff
@@ -0,0 +1,37 @@
...
+    help="Output JSON file name, defaults to end of directory string passed as input",
+)
+def create_config(directory, output):
+    """Create surfactant input configuration file based on input directory."""
+    starting_directory = Path(directory)
+    extract_paths = []
+
+    # Check if there are any files in the directory
+    if any(item.is_file() for item in starting_directory.iterdir()):
+        extract_paths.append(str(starting_directory).replace("\\", "/"))
```
nightlark, line 22:
```suggestion
        extract_paths.append(starting_directory.as_posix())
```

`as_posix()` combines path components using `/` and returns the resulting path as a `str`.

surfactant/cmd/createconfig.py
```diff
@@ -0,0 +1,37 @@
...
+    else:
+        # If there are no files, add the immediate subdirectories
+        for item in starting_directory.iterdir():
+            if item.is_dir():
+                extract_paths.append(str(item).replace("\\", "/"))
```
nightlark, line 27:
```suggestion
                extract_paths.append(item.as_posix())
```
Same as previous change.

surfactant/cmd/createconfig.py
```diff
@@ -0,0 +1,37 @@
+import json
+from pathlib import Path
```
nightlark, line 2:
```suggestion
from loguru import logger
from pathlib import Path
```

surfactant/cmd/createconfig.py
```diff
@@ -0,0 +1,37 @@
...
+
+    config_dict = {"extractPaths": extract_paths, "installPrefix": "/"}
+    config_out = [config_dict]
+
+    output_file_name = output or starting_directory.stem + ".json"
+
+    with open(output_file_name, "w") as json_file:
+        json.dump(config_out, json_file, indent=4)
+
+    click.echo(f"Data written to {output_file_name}")
```
nightlark, line 37:
```suggestion
    logger.info(f"Data written to {output_file_name}")
```
Switch to using logger instead of print.

surfactant/cmd/createconfig.py
```diff
@@ -0,0 +1,37 @@
...
+    # Check if there are any files in the directory
+    if any(item.is_file() for item in starting_directory.iterdir()):
+        extract_paths.append(str(starting_directory).replace("\\", "/"))
+    else:
+        # If there are no files, add the immediate subdirectories
+        for item in starting_directory.iterdir():
+            if item.is_dir():
+                extract_paths.append(str(item).replace("\\", "/"))
+
+    config_dict = {"extractPaths": extract_paths, "installPrefix": "/"}
```
nightlark, line 29: I could see also having an option for a user to specify the install prefix to use.

theStache: added option to specify install prefix, updated readme

surfactant/cmd/createconfig.py
```diff
@@ -0,0 +1,44 @@
...
+    extract_paths = []
+
+    # Check if there are any files in the directory
+    if any(item.is_file() for item in starting_directory.iterdir()):
+        extract_paths.append(starting_directory.as_posix())
+    else:
+        # If there are no files, add the immediate subdirectories
+        for item in starting_directory.iterdir():
+            if item.is_dir():
+                extract_paths.append(item.as_posix())
```
nightlark, line 34: A few questions that came up when I was testing with this set up:

```
.
└── a
    ├── b
    │   └── ls1
    └── c
        └── ls2
```

When I do create-config `a`, I get a config file with `a/b` and `a/c` as the extract paths -- the result when generating an sbom is that any files under both of those extract paths will appear to be installed under the same install prefix (`/ls1` and `/ls2`).

On the other hand, if a file exists as an immediate child of `a`, the extract path will be `a`, and the resulting install paths will be `/b/ls1` and `/c/ls2`.

1. Is that the desired behavior, or would it be more intuitive to have the extract path be `a` regardless of the presence of a file?
2. Would it make sense to use absolute paths in the list of extract paths? The config file wouldn't be portable between different computers, but then `surfactant generate` wouldn't need to be run from the same folder as where the config file got generated.

theStache: It was the desired behavior but I modified to just use the input directory as that makes things more intuitive. Having it use the full paths enables surfactant to be run from any location on the system without issues. With how easy it is is create the config it shouldn't be an issue that its not portable between computers.

surfactant/cmd/createconfig.py
```diff
@@ -0,0 +1,30 @@
...
+    type=str,
+    default="/",
+    help="Install prefix to use in the configuration (default: '/')",
+)
+def create_config(directory, output, install_prefix):
+    """Create surfactant input configuration file based on input directory."""
+    extract_paths = [Path(directory).as_posix()]
+    config_dict = {"extractPaths": extract_paths, "installPrefix": install_prefix}
+    config_out = [config_dict]
+    output_file_name = output or Path(directory).stem + ".json"
```
nightlark, line 27: Some weird cases: if the directory is `.` the default output file name is `.json`, and if the directory is `..` the output name is `...json`. These are probably just situations where a user would want to specify an output name; otherwise, `.json` and `...json` are valid file names and work fine.

nightlark (approved):

PR #57, opened 2023-10-17
Plugin string extractor

theStache:
A plugin for Surfactant that uses the binary2strings library (https://github.com/glmcdona/binary2strings)
Python library to extract strings from ELF and PE files. After installing the plugin it will generate additional json files in the working directory that contain the strings of those files. If there are duplicate hashed files the extractor will skip the entry.
Example:
Output Filename: $(md5hash)_additional_metadata.json
{
"md5hash":"",
"filename":[],
"strings":[]
}

Also added a new metadata_plugin function that is called from inside the get_software_entry, this will be used to add additional plugins that extract information that will be tied to but not included in the SBOM for future analysis efforts.


surfactantplugin-stringextract.py/surfactantplugin_stringextract.py
```diff
@@ -0,0 +1,96 @@
...
+
+@surfactant.plugin.hookimpl
+def extract_strings(filename: str, hash: str, filetype: str, min_len=4):
+    """
+    Extract ASCII strings from a binary file using binary2strings.
+    :param filename (str): The full path to the file to extract information from.
+    :param hash (str): md5 hash of the file
+    :param filetype (str): File type information based on magic bytes.
+    :param min_len (int): Minimum length of strings to be considered valid.
+    """
```
nightlark, line 24: Rather than adding a new hookspec for the new plugins, I think we can reuse the existing `extract_file_info` hook so that a new hookspec doesn't need to get added for each individual plugin. An explicit `return None` could get added, or the existing implicit `return None` will make so that the hook will run, but not create any additional metadata objects for the software entry.

```suggestion
@surfactant.plugin.hookimpl(specname="extract_file_info")
def extract_strings(sbom: SBOM, software: Software, filename: str, filetype: str):
    """
    Extract ASCII strings from a binary file using binary2strings.
    :param sbom(SBOM): The SBOM that the software entry/file is being added to. Can be used to add observations or analysis data.
    :param software(Software): The software entry associated with the file to extract information from.
    :param filename (str): The full path to the file to extract information from.
    :param filetype (str): File type information based on magic bytes.
    :param min_len (int): Minimum length of strings to be considered valid.
    """
    hash = str(software.md5)
    min_len = 4
```

surfactantplugin-stringextract.py/pyproject.toml
```diff
@@ -0,0 +1,33 @@
+[build-system]
+requires = ["setuptools", "setuptools-scm"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "surfactantplugin_stringextract.py"
```
nightlark, line 6: Since this plugin is using the `binary2strings` module, we could consider naming it after that. (For the checksec.py plugin, the `.py` is actually part of the name of the module it uses -- I don't think adding `.py` at the end of the names is a practice we need to follow).

```suggestion
name = "surfactantplugin-binary2strings"
```

surfactantplugin-binary2strings/pyproject.toml
```diff
@@ -0,0 +1,33 @@
+[build-system]
+requires = ["setuptools", "setuptools-scm"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "surfactantplugin_binary2strings.py"
```
nightlark, line 6:
```suggestion
name = "surfactantplugin-binary2strings"
```

nightlark (approved):
One small thing I noticed with the name, otherwise looks good. 👍

PR #58, opened 2023-10-20
Plugin angr

theStache:
A plugin for Surfactant that uses the angr https://github.com/angr/angr Python library to extract the imported function names from ELF and PE files. After installing the plugin, run Surfactant to generate an SBOM as usual and entries for ELF and PE files will generate additional json files in the working directory that contain the list of imported function names of the executable files. If there are duplicate hashed files the extractor will skip the entry.
Example:
`Output Filename: $(md5hash)_additional_metadata.json`

{
  "md5hash": "",
  "filename": [],
  "imported function names": []
}


nightlark (commented):
Thanks for adding these plugins! Sorry it took a while, this and the angr plugin are interesting, and are starting to touch on some aspects of the plugin system that I want to improve support for. I'll mention them at the working group meeting tomorrow, and if you're on the call maybe we can get some discussion going about how to best support them.

A brief summary of some of the things I think it should be able to handle:
* Supporting user config options for plugins
    - options/config file in toml/yaml/json format, that allows user to provide a section for options to give to a plugin
    - flexible command line option that some pattern that indicates what options to pass to a plugin (possibly a single option that takes a json/toml/yaml string to pass to the plugin?)
* Support tying analysis data/observation related files generated by a plugin to a software entry
    - helper functions in SBOM class to add analysis data/observation entry with star relationship to software entry?
    - naming convention for additional files (+ containing folders), including location to create them in, and maybe helper functions to get names to use?

surfactantplugin-angrimportfinder/pyproject.toml
```diff
@@ -0,0 +1,33 @@
+[build-system]
+requires = ["setuptools", "setuptools-scm"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "surfactantplugin_angrimportfinder.py"
```
nightlark, line 6:
```suggestion
name = "surfactantplugin-angrimportfinder"
```

nightlark (approved):

PR #83, opened 2023-11-22
CYT 659 support dynamic list to analyze

Czatar:
- Created a new sbomtype ```ContextEntry``` which holds the three fields in each config entry
- In ```generate.py```, loaded config list into a queue after validating config
- Added context as a parameter to the ```extract_file_info``` hook and passed it in from ```generate.py```

### DIFF
--- a/surfactant/cmd/generate.py
+++ b/surfactant/cmd/generate.py
@@ -19,9 +19,7 @@ from surfactant.sbomtypes import SBOM, Software
 
 
 # Converts from a true path to an install path
-def real_path_to_install_path(
-    root_path: str, install_path: str, filepath: str
-) -> str:
+def real_path_to_install_path(root_path: str, install_path: str, filepath: str) -> str:
     return re.sub("^" + root_path + "/", install_path, filepath)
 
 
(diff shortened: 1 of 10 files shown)


surfactant/cmd/generate.py
Czatar, line 1: I haven't touched a lot of these lines for this PR. Out of curiosity, how did the main branch end up with these longer lines without the pre-commit auto fixes making changes to shorten these lines? Could it be that I'm missing a linter/fixer config file somewhere?

nightlark: The Python code formatting tool used (black) runs on an entire file -- not sure how these lines slipped through, but maybe a recent update to black changed some exceptions to the line length limit (or the files hadn't been updated since the automatic pre-commit check was enabled.

surfactant/sbomtypes/_contextentry.py
```diff
@@ -0,0 +1,13 @@
...
+# SPDX-License-Identifier: MIT
+from dataclasses import dataclass
+from typing import List, Optional
+
+
+@dataclass
+class ContextEntry:
+    extractPaths: List[str]
+    archive: Optional[str] = None
+    installPrefix: Optional[str] = None
```
nightlark, line 13: I'm thinking instead of having this under `sbomtypes`, let's create something like a `surfactant/utilities.py`, `surfactant/helpers.py`, `surfactant/config.py` or `surfactant/context.py` file that can have some useful data types like this that aren't part of the SBOM data, but could be useful in several places.

Something like `config.py` could eventually tie in with providing more user configurable settings and letting plugins define custom options to control their behavior.

surfactant/cmd/generate.py
```diff
@@ -313,8 +318,8 @@ def sbom(
                             # We need get_software_entry to look at the true filepath
                             filepath = true_filepath
 
-                        if install_prefix is not None:
-                            install_path = install_prefix
+                        if entry.installPrefix:
```
nightlark, line 321: I think this one might have been checking against `is not None` so that the install prefix can be set to an empty string. Maybe that intent would be clearer by adding an `or entry.installPrefix == ""` to the condition.

PR #84, opened 2023-12-04
Fix crash with .NET assemblies that have no hex attribute for their PublicKey field

nightlark:
Fixes a corner case with .NET assemblies that have no PublicKey field causing Surfactant to crash due to not having a "hex" attribute/function. This seems to mostly occur due to the PublicKey field being NoneType.


Czatar (commented):
I've attached compiled assemblies and ran surfactant on them. To summarize, I couldn't compile something that breaks surfactant, so I think it might be a very old version (Framework 1.0) that wouldn't include a PublicKey entry in metadata, but I wasn't able to try that. 

[PublicKey.txt](https://github.com/LLNL/Surfactant/files/13639189/PublicKey.txt)
https://drive.google.com/file/d/11KM6-kJzFTaLJtB4EA3eHJUC285SNZWn/view?usp=sharing

nightlark:
This is one of the files that can cause Surfactant to crash (the program this is from has several other files that can also cause a crash, all seem to be localization/resource related).

[ko-KR.zip](https://github.com/LLNL/Surfactant/files/13639502/ko-KR.zip)

Czatar:
Thanks! Here are screenshots of the PublicKey entries under the Assembly Table. In the localization resource file, it looks like the "Absolute Offset" is set to 0x0 and might be related to there not being a Blob section at all, which is where the PublicKey values are stored. I think this might be why dnFile leaves the PublicKey field and other Blob related entries as None?

https://drive.google.com/drive/folders/1dwqBL0bgfri7WO_DIRGNd9Opc7b9xmqU?usp=sharing

nightlark:
Ah, that would make sense. Did you want to try reproducing a file like that for tests? And are there any changes you see that should be made in this PR?

Czatar:
Here's a link to a minimal test case with surfactant output:
https://drive.google.com/drive/folders/1hM-_SRh9Fi394R_g3yTMk1aYeS3m7d2r?usp=sharing

It only really needs ```Example1.exe``` and the ```fr``` directory to run. The files other than ```out.txt``` and ```conf.txt``` are there to show how I compiled it. I followed the instructions in: https://learn.microsoft.com/en-us/dotnet/core/extensions/package-and-deploy-resources

I tried another localization example in:
https://learn.microsoft.com/en-us/dotnet/core/extensions/create-satellite-assemblies
But the main branch handled this case fine. I think it's because the .resource.dll files were properly signed, so the output .resource.dll files contained blobs to store the public key, where the instructions I followed led to an unsigned .resource.dll that does not have a blob.

Since it handles unsigned and signed .resource.dll files just fine, I think that change is good for this PR. Since the dnFile library warns against empty blob tables, I think at some point it's worth looking into which table class attributes are left as None that Surfactant depends on if this happens:
https://github.com/malwarefrank/dnfile/blob/b837ad6381a33f85ee0963cfa50f4ad91609b319/src/dnfile/base.py#L357

Would this be worth its own GitHub issue?

nightlark:
I think it would be worth creating an issue to review use of attributes from dnfile in Surfactant. The value being None isn’t necessarily an issue with most of the information dnfile gets, unless we are trying to call a function (that then doesn’t exist).

Czatar (approved):

PR #106, opened 2023-12-27
Add plugins subfolder

nightlark:
### Summary

If merged this pull request will create a plugins subfolder and move existing plugins to the plugins subfolder. This will hopefully keep things a bit more manageable as additional plugins may get added to the Surfactant repository.

### Proposed changes

- Create plugins folder
- Move existing plugins into the plugins folder


Czatar (approved):
Looks good, and installing/uninstalling a plugin works properly.

PR #119, opened 2024-01-23
Cyt 688 implement surfactant cli find

shaynakapadia:
### Summary

If merged this pull request will add basic functionality to the surfactant cli find command. It will allow users to filter software entries in their SBOM based on a specific file, sha256 hash, installPath and/or containerPath.

### Proposed changes

No changes, just additions to the surfactant cli find command allowing for searches based on a small subset of features.

### Proposed Usage
Searching for all software entries containing "81983020/ " in the containerPath and "User/Bob/Program Files" in the installPath. Results output stdout, but can be redirected to a json file.
`surfactant cli find --containerPath 81983020/ --installPath "User/Bob/Program Files" sample_sbom.json > subset.json`
If no matches, surfactant will log an error. 

Searching for the entry for a specific file /home/file.msi
`surfactant cli find --file /home/file.msi sample_sbom.json > subset.json`

Searching for a specific sha256 hash:
`surfactant cli find --sha256 897498273497293847923 sample_sbom.json > subset.json`

### DIFF
--- a/surfactant/cmd/cli.py
+++ b/surfactant/cmd/cli.py
@@ -6,19 +6,23 @@ from loguru import logger
 
 from surfactant.plugin.manager import find_io_plugin, get_plugin_manager
 from surfactant.sbomtypes._sbom import SBOM
+from surfactant.sbomtypes._software import Software
 
 
 @click.argument("sbom", type=click.File("r"), required=True)
 @click.option("--file", is_flag=False, help="File of the entry to find")
-@click.option("--sha256", is_flag=False, help="sha256 hash of the entry to find")
+@click.option("--sha256", is_flag=False, type=str, help="sha256 hash of the entry to find")
+@click.option("--uuid", is_flag=False, type=str, help="uuid of the entry to find")
 @click.option(
     "--installPath",
     is_flag=False,
+    type=str,
     help="Matches all entries with a install path or partial install path match",
 )
 @click.option(
     "--containerPath",
     is_flag=False,
+    type=str,
     help="Matches all entries with a container path or partial container path match",
 )
 @click.option(
@@ -34,26 +38,16 @@ from surfactant.sbomtypes._sbom import SBOM
     help="SBOM input format, assumes that all input SBOMs being merged have the same format, options=surfactant.input_readers.[cytrics|cyclonedx|spdx]_reader",
 )
 @click.command("find")
-def find(sbom, sha256, file, installpath, containerpath, output_format, input_format):
+def find(sbom, output_format, input_format, **kwargs):
     "CLI command to find specific entry(s) within a supplied SBOM"
     pm = get_plugin_manager()
     output_writer = find_io_plugin(pm, output_format, "write_sbom")
     input_reader = find_io_plugin(pm, input_format, "read_sbom")
     in_sbom = input_reader.read_sbom(sbom)
-    out_sbom = SBOM()
-    if sha256:  # Specific hash, no need to check any other params
-        out_sbom.add_software(in_sbom.find_software(sha256))
-    elif file:  # Specific file referenced, only return that entry, other options not neccessary
-        sha256, sha1, md5 = _calculate_hashes(file)
-        out_sbom.add_software(in_sbom.find_software(sha256))
-    else:
-        path_query = {}
-        if installpath:
-            path_query["installPath"] = str(installpath)
-        if containerpath:
-            path_query["containerPath"] = str(containerpath)
-        for sw in in_sbom.find_software_by_path(path_query):
-            out_sbom.add_software(sw)
+
+    # Remove None values
+    filtered_kwargs = dict({(k, v) for k, v in kwargs.items() if v is not None})
+    out_sbom = cli_find().execute(in_sbom, **filtered_kwargs)
     if not out_sbom.software:
         logger.warning("No software matches found with given parameters.")
     output_writer.write_sbom(out_sbom, sys.stdout)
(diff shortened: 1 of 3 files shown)


shaynakapadia:
Basic functionality working, but marking as draft as I'm still mulling over how I want to do the "finding". Going to figure out a better workflow for applying multiple filters.

surfactant/cmd/cli.py
```diff
@@ -1,10 +1,55 @@
...
 
 @click.argument("sbom", type=click.File("r"), required=True)
+@click.option("--file", is_flag=False, help="File of the entry to find")
+@click.option("--sha256", is_flag=False, type=str, help="sha256 hash of the entry to find")
+@click.option("--uuid", is_flag=False, type=str, help="uuid of the entry to find")
+@click.option(
+    "--installPath",
+    is_flag=False,
+    type=str,
+    help="Matches all entries with a install path or partial install path match",
```
nightlark, line 19:
```suggestion
    help="Matches all entries with an install path or partial install path match",
```

surfactant/cmd/cli.py
```diff
@@ -1,10 +1,55 @@
...
+    "--containerPath",
+    is_flag=False,
+    type=str,
+    help="Matches all entries with a container path or partial container path match",
+)
+@click.option(
+    "--output_format",
+    is_flag=False,
+    default="surfactant.output.cytrics_writer",
+    help="SBOM output format, options=surfactant.output.[cytrics|csv|spdx]_writer",
```
nightlark, line 31: The short names for the plugins should also work for specifying the output format (`cytrics|csv|spdx|cyclonedx`).

surfactant/cmd/cli.py
```diff
@@ -1,10 +1,55 @@
...
+)
+@click.option(
+    "--input_format",
+    is_flag=False,
+    default="surfactant.input_readers.cytrics_reader",
+    help="SBOM input format, assumes that all input SBOMs being merged have the same format, options=surfactant.input_readers.[cytrics|cyclonedx|spdx]_reader",
```
nightlark, line 37: Short names should also work here, though the cyclonedx and spdx readers aren't merged in yet.

surfactant/cmd/cli.py
```diff
@@ -17,3 +62,96 @@ def edit(sbom):
...
     "CLI command to add specific entry(s) to a supplied SBOM"
+
+
+class cli_find:
+    """
+    A class that implements the surfactant cli find functionality
+
+    Attributes:
+    match_functions     A dictionary of functions that provide matching functionality for given SBOM fields (i.e. uuid, sha256, installpath, etc)
+    sbom                A internal record of sbom entries the class adds to as it finds more matches.
```
nightlark, line 73:
```suggestion
    sbom                An internal record of sbom entries the class adds to as it finds more matches.
```

surfactant/cmd/cli.py
```diff
@@ -17,3 +62,96 @@ def edit(sbom):
...
+        for sw in input_sbom.software:
+            match = True
+            for k, v in kwargs.items():
+                if k == "file":
+                    entry_value = sw.sha256
+                    v, _, _ = self._calculate_hashes(v, sha256=True)
+                else:
+                    entry_value = vars(sw)[
+                        self.camel_case_conversions[k] if k in self.camel_case_conversions else k
+                    ]
```
nightlark, line 111: What happens if the input SBOM doesn't have the given key?

surfactant/cmd/cli.py
```diff
@@ -17,3 +62,96 @@ def edit(sbom):
...
+        """Executes the main functionality of the cli_find class
+        param: input_sbom   The sbom to find matches within
+        param: kwargs:      Dictionary of key/value pairs indicating what features to match on
+        """
+        for sw in input_sbom.software:
+            match = True
+            for k, v in kwargs.items():
+                if k == "file":
+                    entry_value = sw.sha256
+                    v, _, _ = self._calculate_hashes(v, sha256=True)
```
nightlark, line 107: This can be a time consuming operation, especially for large files (and repeated for each of possibly thousands of software entries). Can this be changed so that the hash only gets calculated once? Maybe the camel case conversion can also be done so that it doesn't need to happen multiple times.

surfactant/cmd/cli.py
```diff
@@ -17,3 +62,96 @@ def edit(sbom):
...
+    def execute(self, input_sbom: SBOM, **kwargs):
+        """Executes the main functionality of the cli_find class
+        param: input_sbom   The sbom to find matches within
+        param: kwargs:      Dictionary of key/value pairs indicating what features to match on
+        """
+        for sw in input_sbom.software:
+            match = True
+            for k, v in kwargs.items():
+                if k == "file":
+                    entry_value = sw.sha256
```
nightlark, line 106: In theory sha256 could be `None`, since the CyTRICS schema only requires one of the hashes to be present, and other SBOM formats may specify that only sha1 is needed -- in this case, it might be more reliable to calculate all 3 hashes, then require that any that are present in the SBOM match (and that at least one hash is present).

surfactant/cmd/cli.py
```diff
@@ -17,3 +62,96 @@ def edit(sbom):
...
+                    v, _, _ = self._calculate_hashes(v, sha256=True)
+                else:
+                    entry_value = vars(sw)[
+                        self.camel_case_conversions[k] if k in self.camel_case_conversions else k
+                    ]
+                if not self.match_functions[type(entry_value)](entry_value, v):
```
nightlark, line 112: What happens if the entry value decoded from the input SBOM file is a different type?

This is one place that I think the Python 3.10 match..case syntax would be nice to use; but I'm not quite ready to drop support for Python 3.8 and 3.9 yet.

surfactant/cmd/cli.py
shaynakapadia: Theoretically it should always have those keys since the inputted SBOM is of the SBOM class type, and if no value exists for a provided key (which can only be one of the options specified), it shows up as None/Null in the SBOM. In case the other code gets edited, I'll add a key check in there though

surfactant/cmd/cli.py
shaynakapadia: Ah yea good point! Updated to only perform hash calculation once at the beginning

surfactant/cmd/cli.py
shaynakapadia: Since we are matching types from the SBOM class, we should not be getting float, dict or tuples. I've added a handler case for None, but I can add the rest of the types and match them with a generic non implementation function

surfactant/cmd/cli.py
shaynakapadia: Key check added! If key doesn't exist, it will just call the match_none_or_unhandled function

surfactant/cmd/cli.py
shaynakapadia: Updated so it will look for a non zero amount of hash matches

surfactant/cmd/cli.py
shaynakapadia: Updated dic to include values for float, dict, none, and tuple. If a user supplies an SBOM that contains a key with value of that type, it will call the match_none_or_unhandled function

nightlark (approved):

nightlark:
Thanks for adding tests! 💯 🥳

Issue #75, opened 2023-11-06
Explore recognizing JavaScript libraries


nightlark:
[retire.js](https://github.com/retirejs/retire.js/) is a vulnerability scanner for JavaScript, and has a file with a collection of patterns for recognizing various JavaScript libraries and their versions.

1. Look into what it would take to identify a file as JavaScript (or HTML/CSS)
2. Add a method to recognize specific JavaScript libraries and their versions using a database, either from [retire.js](https://github.com/retirejs/retire.js/) or another tool that has a similar collection of patterns that we can use as a starting point for pattern matching.

PR #121, opened 2024-01-27
Categorize web and script files

Czatar:
### Summary

Categorize web files and scripts for later analysis. Related to [issue #75](https://github.com/LLNL/Surfactant/issues/75).

### Proposed changes

Added branches to magic bytes that cover python, bash, html, css, and js. The bash and html branches use magic bytes in case of a missing file type extension.

### DIFF
--- a/surfactant/filetypeid/id_extension.py
+++ b/surfactant/filetypeid/id_extension.py
@@ -6,6 +6,8 @@ import pathlib
 import re
 from typing import Optional
 
+from loguru import logger
+
 import surfactant.plugin
 
 
@@ -34,18 +36,25 @@ def identify_file_type(filepath: str) -> Optional[str]:
     }
     try:
         with open(filepath, "rb") as f:
-            suffix = pathlib.Path(filepath).suffix.lower()
             head = f.read(256)
-            if suffix in _filetype_extensions:
-                return _filetype_extensions[suffix]
             if head.startswith(b"<!DOCTYPE html>"):
                 return "HTML"
             if head.startswith(b"#!") and b"\n" in head:
-                head = head[: head.index(b"\n")].decode("utf-8")
-                for interpreter, filetype in _interpreters.items():
-                    if re.search(interpreter, head):
-                        return filetype
+                end_line = head.index(b"\n")
+                try:
+                    head = head[:end_line].decode("utf-8")
+                    for interpreter, filetype in _interpreters.items():
+                        if re.search(interpreter, head):
+                            return filetype
+                except UnicodeDecodeError:
+                    logger.warning(
+                        f"Decode error in file {filepath}: {head[: end_line]}"
+                    )
                 return "SHEBANG"
-            return None
     except FileNotFoundError:
+        logger.warning(f"File not found: {filepath}")
         return None
+    suffix = pathlib.Path(filepath).suffix.lower()
+    if suffix in _filetype_extensions:
+        return _filetype_extensions[suffix]
+    return None
--- a/surfactant/filetypeid/id_extension.py
+++ b/surfactant/filetypeid/id_extension.py
@@ -38,7 +38,7 @@ def identify_file_type(filepath: str) -> Optional[str]:
             head = f.read(256)
             if suffix in _filetype_extensions:
                 return _filetype_extensions[suffix]
-            if head[:14] == b"<!DOCTYPE html>":
+            if head.startswith(b"<!DOCTYPE html>"):
                 return "HTML"
             if head.startswith(b"#!") and b"\n" in head:
                 head = head[: head.index(b"\n")].decode("utf-8")
--- a/surfactant/filetypeid/id_extension.py
+++ b/surfactant/filetypeid/id_extension.py
@@ -12,27 +12,39 @@ import surfactant.plugin
 @surfactant.plugin.hookimpl
 def identify_file_type(filepath: str) -> Optional[str]:
     # pylint: disable=too-many-return-statements
+    _filetype_extensions = {
+        ".sh": "SHELL",
+        ".bash": "BASH",
+        ".zsh": "ZSH",
+        ".py": "PYTHON",
+        ".pyc": "PYTHON_COMPILED",
+        ".js": "JAVASCRIPT",
+        ".css": "CSS",
+        ".html": "HTML",
+        ".htm": "HTML",
+        ".php": "PHP",
+    }
+    _interpreters = {
+        "sh": "SHELL",
+        "bash": "BASH",
+        "zsh": "ZSH",
+        "php": "PHP",
+        "python": "PYTHON",
+        "python3": "PYTHON",
+    }
     try:
-        with open(filepath, "r") as f:
-            head = f.readline().strip("\n")
+        with open(filepath, "rb") as f:
             suffix = pathlib.Path(filepath).suffix.lower()
-            # Check for script files
-            if re.match(r"#!.*bash", head) or suffix == ".sh":
-                return "BASH"
-            if re.match(r"#!.*zsh", head) or suffix == ".zsh":
-                return "ZSH"
-            if re.match(r"#!.*php", head) or suffix == ".php":
-                return "PHP"
-            if re.match(r"#!.*python", head) or suffix == ".py":
-                return "PYTHON"
-            if suffix == ".pyc":
-                return "PYTHON_COMPILED"
-            if suffix == ".js":
-                return "JAVASCRIPT"
-            if suffix == ".css":
-                return "CSS"
-            if head == "<!DOCTYPE html>" or suffix in (".html", ".htm"):
+            head = f.read(256)
+            if suffix in _filetype_extensions:
+                return _filetype_extensions[suffix]
+            if head[:14] == b"<!DOCTYPE html>":
                 return "HTML"
+            if head.startswith(b"#!") and b"\n" in head:
+                head = head[: head.index(b"\n")].decode("utf-8")
+                for interpreter, filetype in _interpreters.items():
+                    if re.search(interpreter, head):
+                        return filetype
             return None
     except FileNotFoundError:
         return None
--- a/surfactant/filetypeid/id_extension.py
+++ b/surfactant/filetypeid/id_extension.py
@@ -47,9 +47,7 @@ def identify_file_type(filepath: str) -> Optional[str]:
                         if re.search(interpreter, head):
                             return filetype
                 except UnicodeDecodeError:
-                    logger.warning(
-                        f"Decode error in file {filepath}: {head[: end_line]}"
-                    )
+                    logger.warning(f"Decode error in file {filepath}: {head[: end_line]}")
                 return "SHEBANG"
     except FileNotFoundError:
         logger.warning(f"File not found: {filepath}")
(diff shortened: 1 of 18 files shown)


Czatar:
Wondering if:
- File type extension should be made into its own function or if it's fine to have it where it is
- There are any file types to include or exclude for this PR

surfactant/filetypeid/id_magic.py
```diff
@@ -162,6 +162,18 @@ def identify_file_type(filepath: str) -> Optional[str]:
...
+            if magic_bytes[:12] == b"#!/bin/bash\n" or suffix == ".sh":
+                return "BASH"
+            if suffix == ".py":
+                return "PYTHON"
+            # Check for web related files
+            if suffix == ".js":
+                return "JAVASCRIPT"
+            if suffix == ".css":
+                return "CSS"
+            if magic_bytes[:16] == "<!DOCTYPE html>\n" or suffix == ".html":
```
nightlark, line 175: `.htm` is another common extension for HTML files. I think syntax `suffix in (".html", ".htm")` can be used to avoid a long list of `or` conditions.

surfactant/filetypeid/id_magic.py
```diff
@@ -162,6 +162,18 @@ def identify_file_type(filepath: str) -> Optional[str]:
                 int.from_bytes(magic_bytes[0:4], byteorder="big", signed=False) & 0xFF0F80FF
             ) == 0xF00D0000:
                 return "OMF_LIB"
+            # Check for script files
+            if magic_bytes[:12] == b"#!/bin/bash\n" or suffix == ".sh":
+                return "BASH"
+            if suffix == ".py":
```
nightlark, line 168: I think `#!/bin/python` is one of the things that can appear at the top of some Python files without a `.py` extension -- check the GE G500 python scripts to confirm that's what it uses (or ask Alex Armstrong).

I think we can also add a check for a `.pyc` extension, that returns something like `PYTHON_COMPILED` or `PYTHON_BYTECODE`?

nightlark:
>     * File type extension should be made into its own function or if it's fine to have it where it is

I was thinking it could be made into its own function, since it isn't really a check based on magic bytes. Using the `#!` line is also maybe more of a heuristic than magic bytes?

surfactant/filetypeid/id_extension.py
```diff
@@ -0,0 +1,38 @@
...
+            head = f.readline().strip("\n")
+            suffix = pathlib.Path(filepath).suffix.lower()
+            # Check for script files
+            if re.match(r"#!.*bash", head) or suffix == ".sh":
+                return "BASH"
+            if re.match(r"#!.*zsh", head) or suffix == ".zsh":
+                return "ZSH"
+            if re.match(r"#!.*php", head) or suffix == ".php":
+                return "PHP"
+            if re.match(r"#!.*python", head) or suffix == ".py":
```
Czatar, line 26: I saw [this post](https://stackoverflow.com/questions/2429511/why-do-people-write-usr-bin-env-python-on-the-first-line-of-a-python-script) and I think it might be a good idea to match just the script interpreter name in case the path is different across projects. Would this be fine or should it still include "/bin/" or "/usr/bin/env/" since it's very likely to follow this pattern anyways?

nightlark: Hm... okay, so I'm looking at things a bit more -- one common line is the shebang for running `/usr/bin/env bash` (note lack of a trailing slash after env, it's a program) for portability on different OSes.

What I'm thinking right now is, let's restructure this check slightly to split out the suffix checks into separate if statements, and then add a check for `head.startswith("#!")`, and then within that have a nested set of checks for what program is being run, with a default case that returns something like "SHEBANG".

surfactant/filetypeid/id_extension.py
```diff
@@ -0,0 +1,38 @@
...
+
+@surfactant.plugin.hookimpl
+def identify_file_type(filepath: str) -> Optional[str]:
+    # pylint: disable=too-many-return-statements
+    try:
+        with open(filepath, "r") as f:
+            head = f.readline().strip("\n")
+            suffix = pathlib.Path(filepath).suffix.lower()
+            # Check for script files
+            if re.match(r"#!.*bash", head) or suffix == ".sh":
```
nightlark, line 20: Two things here: for the `#!` check, `sh` should be recognized (probably) as `SHELL`, and same for a `.sh` extension (I've seen some file suffixes of `.bash` for bash scripts, but otherwise, the `.sh` doesn't indicate what interpreter should be used.

surfactant/filetypeid/id_extension.py
nightlark: Another thing is for the `#!` check, `python3` should also be recognized as `PYTHON`.

nightlark:
Something I wanted to make a note on regarding the regex checks -- I came across a few `#!` lines in a stackoverflow post:

```
#!/usr/bin/env bash
#!/bin/bash
#!/bin/sh
#!/bin/sh -
```

In particular, the last one stands out where the end of the line is passing a `-` as an argument to `sh`. I think in theory the shebang line could have any set of arguments passed to the given command, so to recognize some corner cases it might be necessary to split the line similar to how the OS would to identify the command to run (+ a special case for using the first argument if env is the command).

This sounds like a lot of work though for an uncommon case, so I'm leaning towards adding that as a GitHub issue for the backlog.

surfactant/filetypeid/id_extension.py
```diff
@@ -0,0 +1,38 @@
...
+
+import surfactant.plugin
+
+
+@surfactant.plugin.hookimpl
+def identify_file_type(filepath: str) -> Optional[str]:
+    # pylint: disable=too-many-return-statements
+    try:
+        with open(filepath, "r") as f:
+            head = f.readline().strip("\n")
```
nightlark, line 17: I generated some SBOMs using the changes in this branch -- one issue I encountered is for binary files, Python will throw an exception (UnicodeDecodeError) if there are invalid utf-8 bytes. I think an initial check for just the "#!" starting a file (I suppose these are actually a human readable form of magic bytes) would address most of these cases, but we may want to look for the end of line `\n` byte by looking through the raw bytes first, before trying to decode them as a string.

Czatar:
The issue with having arguments in the shebang line is tricky. From what I know:
- It's possible for the path to the binary to have spaces, so we can't just split by spaces
- It's possible for the arguments to have paths, so just splitting it by slashes also won't help

My next thought would be to check for single/double quotes or backslashes that include the space in the path. Windows doesn't acknowledge shebang lines, so maybe backslash checking won't be a huge issue. For the GitHub issue, Is there anything else that would complicate this?

nightlark (commented):
Forgot to make these comments visible.

surfactant/filetypeid/id_extension.py
```diff
@@ -0,0 +1,55 @@
...
+    try:
+        with open(filepath, "rb") as f:
+            head = f.read(256)
+            if head.startswith(b"<!DOCTYPE html>"):
+                return "HTML"
+            if head.startswith(b"#!") and b"\n" in head:
+                end_line = head.index(b"\n")
+                head = head[:end_line]
+                for interpreter, filetype in _interpreters.items():
+                    if re.search(interpreter, head):
```
nightlark, line 46: Any idea how the performance is for `re.search` vs just using the `in` keyword?

nightlark: Okay, so I put together some quick scripts to test with `time python3 <script>`. I got around 2.3s using `re.search`, and 1.2s using `in`.

```python
import re

interpreters = {
    b"sh": "SHELL",
    b"bash": "BASH",
    b"zsh": "ZSH",
    b"php": "PHP",
    b"python": "PYTHON",
    b"python3": "PYTHON",
}

candidate = b"#!/bin/fsdafsdf/local/bas/bas/bas/bas/pytho/python3"

for i in range(0, 1000000):
    for interpreter, filetype in interpreters.items():
        if re.search(interpreter, candidate):
            pass

```

```python
interpreters = {
    b"sh": "SHELL",
    b"bash": "BASH",
    b"zsh": "ZSH",
    b"php": "PHP",
    b"python": "PYTHON",
    b"python3": "PYTHON",
}

candidate = b"#!/bin/fsdafsdf/local/bas/bas/bas/bas/pytho/python3"

for i in range(0, 1000000):
    for interpreter, filetype in interpreters.items():
        if interpreter in candidate:
            pass
```

Out of curiosity, I tried out using Aho-Corasick (on strings though... using the pyahocorasick library) to see how fast it could be if we really wanted to get a faster search; with this, the test took 0.35s to run:

```python
import ahocorasick

automaton = ahocorasick.Automaton()

interpreters = {
        "sh": "SHELL",
        "bash": "BASH",
        "zsh": "ZSH",
        "php": "PHP",
        "python": "PYTHON",
        "python3": "PYTHON3",
}

for interpreter, filetype in interpreters.items():
    automaton.add_word(interpreter, filetype)

automaton.make_automaton()

candidate = "#!/bin/fsdafsdf/local/bas/bas/bas/bas/pytho/python3"

for i in range(0, 1000000):
    for item in automaton.iter_long(candidate):
        pass
```

nightlark: At some point, it might be nice to switch some of the different file type ID checks over to a faster check -- but saving a second or two isn't that much compared to the time it takes to generate hashes and analyze large binaries.

nightlark (approved):

surfactant/filetypeid/id_extension.py
nightlark: Okay, so I put together some quick scripts to test with `time python3 <script>`. I got around 2.3s using `re.search`, and 1.2s using `in`.

```python
import re

interpreters = {
    b"sh": "SHELL",
    b"bash": "BASH",
    b"zsh": "ZSH",
    b"php": "PHP",
    b"python": "PYTHON",
    b"python3": "PYTHON",
}

candidate = b"#!/bin/fsdafsdf/local/bas/bas/bas/bas/pytho/python3"

for i in range(0, 1000000):
    for interpreter, filetype in interpreters.items():
        if re.search(interpreter, candidate):
            pass

```

```python
interpreters = {
    b"sh": "SHELL",
    b"bash": "BASH",
    b"zsh": "ZSH",
    b"php": "PHP",
    b"python": "PYTHON",
    b"python3": "PYTHON",
}

candidate = b"#!/bin/fsdafsdf/local/bas/bas/bas/bas/pytho/python3"

for i in range(0, 1000000):
    for interpreter, filetype in interpreters.items():
        if interpreter in candidate:
            pass
```

Out of curiosity, I tried out using Aho-Corasick (on strings though... using the pyahocorasick library) to see how fast it could be if we really wanted to get a faster search; with this, the test took 0.35s to run:

```python
import ahocorasick

automaton = ahocorasick.Automaton()

interpreters = {
        "sh": "SHELL",
        "bash": "BASH",
        "zsh": "ZSH",
        "php": "PHP",
        "python": "PYTHON",
        "python3": "PYTHON3",
}

for interpreter, filetype in interpreters.items():
    automaton.add_word(interpreter, filetype)

automaton.make_automaton()

candidate = "#!/bin/fsdafsdf/local/bas/bas/bas/bas/pytho/python3"

for i in range(0, 1000000):
    for item in automaton.iter_long(candidate):
        pass
```

At some point, it might be nice to switch some of the different file type ID checks over to a faster method (magic bytes, srec/hex, etc) -- but saving a second or two isn't that much compared to the time it takes to generate hashes and analyze large binaries.

PR #124, opened 2024-02-03
Replace deprecated setuptools_scm write_to option with version_file

nightlark:
The `write_tool` configuration option in setuptools_scm is deprecated -- `version_file` should be used instead.

See docs for details: https://setuptools-scm.readthedocs.io/en/latest/config/#setuptools_scm.Configuration


Czatar (approved):
Checked with setuptools_scm and didn't find anything else in Surfactant that's deprecated from this. Surfactant still runs fine locally as well.

Issue #130, opened 2024-02-06
Container path may be missing path components


nightlark:
When creating an SBOM from pieces of a container, the files may appear to be in a flat file system within the container.

e.g. `A.zip` has a file under `usr/bin/fileA`

An SBOM is generated with a config file that lists:
archive = "A.zip"
extractPaths = "some-directory-with-only-usr/bin-files"

The resulting SBOM should have fileA with a container path of `<parent_uuid>/usr/bin/fileA`, but it actually just has `<parent_uuid>/fileA`. This could potentially be inferred to some degree from the provided installPrefix? But msi/exe installers get complicated (are they just a bunch of blobs of file data internally?).

nightlark:
Rebasing (absolute?) symlinks has issues in a similar situation.

e.g. `symlnkA -> /usr/bin/fileA`with an extract path with only files in `usr/bin` that doesn't also replicate the `usr` and `bin folders within it results in these lines messing up:

```python
# Rebase to get the true location
current_path = os.path.join(extract_dir, dest)
```

current_path becomes something like `<extract_path>/usr/bin/fileA`... which doesn't exist because it is just `<extract_path>/fileA` on the system that is doing the analysis.

A possible fix could be assuming that the layout for the extracted files and created symlinks were pulled directly from the system they run on, and that `installPrefix` can be used to pull out the part of the path that should not be replicated.

nightlark:
OK, the fix for the symlink resolving works -- however, if files from the same folder (e.g. /usr/bin) (or separate folders such as /bin or /sbin) have been split up into multiple extractPaths, the resolver doesn't consider files in those other directories.

The only way I could see to fix this would be to internally create a "file system mapping" of which installed directory different extract paths map to (one to many). Which gets into more complex questions about how to do SBOMs with files from different systems.

nightlark:
A "quick" solution to address this could be adding a `containerPath` option that specifies a path that should go between the container UUID and the relative paths for files within the extractPaths. Or users can just replicate the directory structure they want it to have in the container by having some nested folders.

I think a separate issue for refactoring Surfactant to use data structure that maps better to a file system hierarchy, and adds a mechanism to the config file to specify a unique ID representing the filesystem a particular config file entry "installs to" will be needed -- if none is specified, everything would be placed under a default filesystem ID. Resolving relationships would then be given a FS tree/hierarchy to consider, and that would avoid establishing links to files not on the same filesystem.

PR #131, opened 2024-02-07
Symlink resolver and get_software_entry return value change related bugfixes

nightlark:
### Summary

If merged this pull request will fix several issues related to resolving symlinks, and the get_software_entry returning a tuple that needs unpacking when used to get an entry for an "archive" file. Some of the things fixed are described some in #130.

The most complex bug involved symlink resolution, when the extractpath corresponds to a subfolder on the system that the files came from (e.g. `C:/Squashfs-usrbin/`corresponds to an install prefix of `/usr/bin`). During symlink resolution, this would result in a symlink such as `fileA -> /usr/bin/fileB` getting resolved to `C:/Squashfs-usrbin/usr/bin/fileB`, whereas it should actually be `C:/Squashfs-usrbin` since `C:/Squashfs-usrbin`==`/usr/bin` on the original file system.

One gotcha that still exists is if the symlink points to either a directory outside `/usr/bin` such as `/bin` or `/sbin` (or multiple extract paths correspond to the same install prefix), the symlink resolver doesn't know about any file system trees outside of the current extract path folder being looked at. Addressing this gets complicated though, and ties in with SBOMs created from files that come from completely different systems.

### Proposed changes

- Unpack tuple from get_software_entry when processing the "archive" file listed in the config entry
- Look-up files in file_symlinks dictionary using sw_parent variable, rather than the no longer existing `entry` variable
- Use installPrefix when resolving symlinks to prevent install prefix related directories from getting unnecessarily appended to the extract path
- Output which install paths are symlinks in the metadata to help with debugging and future analysis

### DIFF
--- a/surfactant/cmd/generate.py
+++ b/surfactant/cmd/generate.py
@@ -138,7 +138,7 @@ def warn_if_hash_collision(soft1: Optional[Software], soft2: Optional[Software])
         elif soft1.size != soft2.size:
             collision = True
     if collision:
-        logger.warn(
+        logger.warning(
             f"Hash collision between {soft1.name} and {soft2.name}; unexpected results may occur"
         )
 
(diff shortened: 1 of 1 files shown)


nightlark:
Thanks for adding tests! :100:

PR #136, opened 2024-02-12
Cyt 640 web UI

KendallHarterAtWork:
### Summary

If merged this pull request will add a web UI for running Surfactant.

### Proposed changes

The UI runs on top of other Surfactant code and does not modify any underlying functionality.  It provides a simple UI for running generate and merge commands using a web GUI.


nightlark:
I'll try to find some time to review and try out these changes either this coming week if I can find some time and have decent internet access for a while -- otherwise I'll get it done March 11.

surfactant/cmd/ui.py
```diff
@@ -0,0 +1,122 @@
...
+                args.append(req["output_format"])
+            merge_command.merge_command(args, standalone_mode=False)
+            return {"error": False, "file_name": output_file_name}
+    except Exception as e:
+        return {"error": True, "error_desc": str(e)}
+
+
+@site.route("/")
+def index():
+    return flask.send_file("../../web-files/ui.html")
```
nightlark, line 111: It looks like loading the html file from this relative path only works when using an editable pip install of Surfactant (from a clone of the git repo). When running from a non-editable pip install (roughly equivalent to `python -m build` and then installing from the built wheel), the `web-files/ui.html` file can't be found, and isn't included in the wheel file created.

I think the solution is moving the `web-files` folder into the `surfactant` folder, and then using a path relative to the `cmd/ui.py` file for locating the html file.

```suggestion
    return flask.send_file(os.path.join(os.path.dirname(__file__), "../web-files/ui.html"))
```

surfactant/cmd/ui.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+@click.command("ui")
+@click.argument("port", type=click.INT, required=False, default=8080)
+def ui(port: int):
+    os.makedirs("results", exist_ok=True)
```
nightlark, line 117:
```suggestion
    """Start web UI server

    An optional PORT (default=8080) can be supplied as an argument.
    """
    os.makedirs("results", exist_ok=True)
```

This should make so the surfactant help message shows a description of the subcommand.

surfactant/cmd/ui.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+@site.route("/")
+def index():
+    return flask.send_file("../../web-files/ui.html")
+
+
+@click.command("ui")
+@click.argument("port", type=click.INT, required=False, default=8080)
+def ui(port: int):
```
nightlark, line 116:
```suggestion
def ui(port: int):
    """Start web UI server

    An optional PORT (default=8080) can be supplied as an argument.
    """
```

This should make so the surfactant help message shows a description of the subcommand.

surfactant/cmd/ui.py
```diff
@@ -0,0 +1,122 @@
...
+    )
+
+
+@site.post("/generate")
+def generate():
+    # Hush some pylint things due to them not really being aplicable here
+    # pylint: disable=broad-exception-caught, no-value-for-parameter
+    try:
+        req = flask.request.get_json()
+        with tempfile.NamedTemporaryFile("w") as config_file:
```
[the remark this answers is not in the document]
KendallHarterAtWork: Yes, the code effectively calls the function with the command line arguments so it needs a file name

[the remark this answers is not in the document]
nightlark: I wonder if in a later PR we should restructure the generate subcommand function in a way that makes so users calling surfactant as a library (`import surfactant.cmd.generate; surfactant.cmd.generate.sbom(...)`) can give it a config file directly in memory.

nightlark (approved):
Two ideas for potential changes to Surfactant:
* Restructure how generate subcommand is organized so users calling Surfactant as a library don't need to write any files to disk
* Determine if there is a typical folder (likely different for macOS/Linux/Windows) used for user configuration and data storage -- this could be used for storing user plugin/setting preferences, databases used by plugins for matching files to packages, and in some cases output result files

PR #140, opened 2024-02-17
Switch to ruff for several pre-commit checks

nightlark:
### Summary

If merged this pull request will replace psf/black, isort, and flake8 (+flake8-bugbear) with [ruff](https://github.com/astral-sh/ruff) for the pre-commit checks. There are minor formatting changes with ruff -- likely a change that was in preview for psf/black and recently moved to stable.

For a time comparison when I tested this change on my laptop, using ruff is 10-16x faster than the replaced tools (.04 seconds to check all files). That said... pylint takes 3-4s to run on all files, which is... slow -- ruff is gradually adding those checks, but has a ways to go (with its pylint equivalent rules enabled, it still looks like it will take well under a second to do the same checks).

### Proposed changes

- Replace black, isort, and flake8 pre-commit checks  with ruff
- Update files that have minor formatting changes when using ruff
- gitignore vscode directory

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,3 +1,0 @@
-[flake8]
-max-line-length = 100
-extend-ignore = E203,E402,E501,F401,F841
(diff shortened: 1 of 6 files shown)


shaynakapadia (approved):
Looks good! Tested it locally with no issues.

PR #158, opened 2024-03-18
Javascript library detection

### LINKED ISSUE #75: Explore recognizing JavaScript libraries

Czatar:
### Summary

If merged this pull request will add support for Javascript file detection. If the file belongs to a Javascript library, it will identify which library and its version number.

### Proposed changes

Adds an info extractor and scripts to update the regex file used to identify library name and version.

Resolves #75


Czatar:
Here's the output that I tested on a sample javascript file:
[out.txt](https://github.com/LLNL/Surfactant/files/14641892/out.txt)

The snippet below is part of the output file that this PR would add to SBOM outputs. Should some of this get renamed to "Javascript_Library"/"Library_Version" or something like that? Or is this fine?
```
{
          "Library": "select2",
          "Version": "3.5.4"
}
```

The other question is where to put the .regex file, or if it should have another extension, since I think .json gets gitignored

surfactant/infoextractors/js_file.py
```diff
@@ -0,0 +1,68 @@
...
+
+def match_by_attribute(attribute: str, content: str, database: Dict) -> Tuple[str, str]:
+    for name, library in database.items():
+        if attribute in library:
+            for pattern in library[attribute]:
+                matches = re.search(pattern, content)
+                if matches:
+                    if len(matches.groups()) > 0:
+                        return (matches.group(1), name)
+                    return ("None Found", name)
```
nightlark, line 67: So what I'm wondering is if for filecontent matches we should skip the rest of the regexes for the current library being checked but then continue on and see if there are matches for any other js libraries based on contents. I'm not sure what the odds of someone creating a single amalgamated json file of all 3rd party json dependencies are for the javascript we are likely to encounter.

scripts/js_libraries/README.md
```diff
@@ -0,0 +1,18 @@
...
+
+## get_retirejs_db.py
+
+Running this script retrieves vulnerability information from various javascript libraries from
+[retire.js](https://github.com/RetireJS/retire.js/blob/master/repository/jsrepository-master.json)
+and only keeps the contents under each library's "extractors" section, which contains regexes
+relevant for detecting a specific javascript library by its file name, file contents, or md5 hash.
+
+The resulting smaller json is written to `reduced.json` in the same directory. This smaller file
+will be read from to make the checks later on.
```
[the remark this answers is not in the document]
nightlark: Since `js_libraries.regex` is a JSON file, maybe we should rename it in a way that has a json extension (e.g. `js_library_patterns.json` or `js_library_regexes_db.json`). And then update any references to the new file name.

nightlark (commented):
One of the things I've been thinking about for this is whether we should include the JSON file of js library regexes in the repository, or if we should do something like download/create the database file on the first run. It looks like the file is fairly small, so let's leave it as-is for now -- some future features that use a similar set of techniques might have much larger databases, at which point we can look into adding some mechanism for downloading/generating the databases on first run + updating them independent of new Surfactant releases on PyPI.

Two other interesting techniques that retire.js uses (to keep in mind for future Surfactant features) are running libraries in a Chrome sandbox (basically a VM), and parsing the javascript code to an AST and then looking for a match in the AST (new feature in retire.js v5).

Before this PR gets merged, we should also make sure to update the js library regexes file.

surfactant/infoextractors/js_file.py
```diff
@@ -0,0 +1,66 @@
...
+def match_by_attribute(attribute: str, content: str, database: Dict) -> List[Dict]:
+    libs = []
+    for name, library in database.items():
+        if attribute in library:
+            for pattern in library[attribute]:
+                matches = re.search(pattern, content)
+                if matches:
+                    if len(matches.groups()) > 0:
+                        libs.append({"Library": name, "Version": matches.group(1)})
+                        break
```
nightlark, line 65:
```suggestion
                        # skip remaining patterns, move on to the next library
                        break
```

Comment to clarify behavior -- in case a future reader misses the nested for loop.

Czatar:
As it is right now, if no libraries are detected, the output is
```json
{
      "jsLibraries": []
}
```
Otherwise, it's a list of detected libraries
```json
{
    "jsLibraries": [
        {
          "Library": "select2",
          "Version": "3.5.4"
        }
    ]
}

File: surfactant/infoextractors/js_file.py [hunk 4]
```diff
@@ -0,0 +1,67 @@
...
+
+def match_by_attribute(attribute: str, content: str, database: Dict) -> List[Dict]:
+    libs = []
+    for name, library in database.items():
+        if attribute in library:
+            for pattern in library[attribute]:
+                matches = re.search(pattern, content)
+                if matches:
+                    if len(matches.groups()) > 0:
+                        libs.append({"Library": name, "Version": matches.group(1)})
```
nightlark, line 64:
```suggestion
                        libs.append({"library": name, "version": matches.group(1)})
```

Making these keys lowercase would more closely match the capitalization used by other keys in the output SBOM.

nightlark:
I tested the changes on a combined js file with both jQuery and Axios in it -- matching multiple libraries in the same JS file worked.

edalily:
Tested with js files with both axios and lodash and detection worked

edalily (approved):
Tested with js files with both axios and lodash and detection worked

PR #167, opened 2024-03-27
Fix bug with iterating through and adding containers to CycloneDX SBOM

nightlark:
Found a bug with outputting SBOM in CycloneDX format -- the function returns a tuple rather than a list, so it can't be iterated over with a for loop. The fix moves the function call higher up so that it can unpack the tuple first, then iterate over the contents of the list that belongs to the tuple.

(I think this should be possible to replicate by generating an SBOM using any of the "HELICS" examples that specify an "archive" so that a "Contains" relationship will be created in the resulting SBOM, and then running surfactant generate with the `--output_format cyclonedx`.)


mcutshaw:
Code looks good, going to try to replicate quick though.

mcutshaw (approved):
I was able to replicate the issue and confirm the pull fixes it (tested with Helics 3.5.2).

Issue #155, opened 2024-03-07
Performance improvements for large filesystems


mcutshaw:
**Is your feature request related to a problem? Please describe.**
We have been having some performance issues with "very large" filesystems > 1 million files.

**Describe the solution you'd like**
A lot of these performance issues seem to stem from iterating through lists that contain each software or relationship entry. If we use Dicts or Sets where applicable we will be reducing the cost of those lookup operations from O(n) to amortized O(1). 

**Describe alternatives you've considered**
None

**Additional context**
Please see https://github.com/mcutshaw/Surfactant/tree/perf_improvements for some possible adjustments. With one filesystem we've tested with it reduced a run that took several hours to less then 5 minutes.

nightlark:
With the changes in the perf_improvements branch, does (de)serializing the dataclasses to/from JSON still result in the same output? The one thing to be careful of is changing the data classes in a way that results in a (CyTRICS) SBOM being output that doesn't pass validation with the (CyTRICS) schema.

If the java file changes make a difference, I'd merge a PR with that change now.

I'm not sure the early return for the find functions is valid -- if I'm understanding the logic, it looks like it will make the search exit as soon as the first non-matching entry is encountered, without considering that any subsequent entries in the various lists could be the one that actually matches; I think the only time the correct result would be returned in this scenario would be if the matching entry is the first one checked.

---

I've been mulling over decoupling the internal BOM data structures used by Surfactant from the CyTRICS schema -- the current set up definitely makes reading/writing CyTRICS SBOMs easy (json.[load|dump]), but makes other tasks really difficult.

Establishing relationships between files based on the filesystem structure would lend itself well to internally storing things as a graph representation, and should make most of the relationship establishing algorithms O(nlogn) instead of O(n^2) -- from a recently run, establishing relationships for a large number of files took most of the time. It would also make so we could just use graph edges to represent relationships between files, without the kind of awkward relationship lists.

I think this structure would still let us have a lookup table by hash for software entries. Or at least a set to tracks which hashes have been seen (md5, sha1, sha256), since ideally the common case is an entry not already existing with a matching hash.

Thoughts or comments?

mcutshaw:
You are correct. On a second look I misunderstood that we were directly serializing the dataclass to JSON. I do believe that the lookup table provided the greatest performance improvements of the changes, so I would prefer to discuss ways to implement that and potentially just tack on the java changes as well.

[...]

PR #168, opened 2024-03-29
Perf2

mcutshaw:
### Summary

If merged this pull request will improve performance and is discussed in #155 

### Proposed changes

See #155

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,18 @@
+# top-most EditorConfig file
+root = true
+
+# Unix-style newlines
+[*]
+end_of_line = lf
+
+# 4 space indentation
+[*.py]
+indent_style = space
+indent_size = 4
+trim_trailing_whitespace = true
+charset = utf-8
+
+# Indentation for YAML files for GitHub Actions workflows
+[.github/workflows/*.yml]
+indent_style = space
+indent_size = 2
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,7 +1,7 @@
 exclude: ^(.gitignore|generate_sbom.py|extract_file_info.py|pe_info.py)
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.3.3
+    rev: v0.3.4
     hooks:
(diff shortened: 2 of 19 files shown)


mcutshaw:
Please note I did disable pylints R0902, let me know if we want to change something as well. It would also appear I broke some of the unit tests, so I will need to fix those as well.

mcutshaw:
@nightlark I believe this one is ready for review.

nightlark (approved):

nightlark (commented):
Actually, one question -- does https://github.com/LLNL/Surfactant/blob/main/surfactant/cmd/merge.py for the merge subcommand need some changes to work with the sets (append -> add) and maintain the sha256 lookup table?

mcutshaw:
> Actually, one question -- does https://github.com/LLNL/Surfactant/blob/main/surfactant/cmd/merge.py for the merge subcommand need some changes to work with the sets (append -> add) and maintain the sha256 lookup table?

Yes, I think there is only one change needed on line 61, but I will make sure and make edits tomorrow.

nightlark:
I was thinking line 61, and two things in https://github.com/LLNL/Surfactant/blob/e2036bb62febeaa5c71913b2acac5fd5f2b6070f/surfactant/sbomtypes/_sbom.py#L120
- a call to relationships.append() in the merge function
- not updating the internal sha256 hashmap with software entries added via the merge function (if the entry to be added has a sha256 that isn’t None)
- the add_software function above the merge function in the SBOM class maybe should also update the hashmap (if the software entry has a sha256 hash)

I wonder if maybe the reason it didn’t break is that the merge function mostly uses the “internal” _ find functions that aren’t using the hashmap?

nightlark (approved):

nightlark:
We might want to consider updating the [input reader for the cytrics format](https://github.com/LLNL/Surfactant/blob/main/surfactant/input_readers/cytrics_reader.py) to rebuild the sha256 hashmap when loading a new SBOM. I don't think this affects any of the functionality built-into Surfactant though, since reading existing SBOMs isn't used much (and where it is, the sha256 lookup table is not used).

PR #174, opened 2024-04-11
Add EditorConfig to give IDEs hints at how to format various file types

### LINKED ISSUE #142: Add EditorConfig file to communicate default settings to IDEs
Add an EditorConfig file so that regardless of IDE used, some default settings like line endings are selected: https://editorconfig.org/

nightlark:
This adds an [EditorConfig](https://editorconfig.org/) which is recognized by a variety of common IDEs used for software development (including vim), and tells them some basic information on how the files should be formatted.

Resolves #142.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,18 @@
+# top-most EditorConfig file
+root = true
+
+# Unix-style newlines
+[*]
+end_of_line = lf
+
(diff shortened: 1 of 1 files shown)


Czatar (approved):
Checked with my text editors and it looks like it's working 👍

PR #176, opened 2024-04-11
Fix install prefix containing Windows-style backslash path separator and catch JSON decode errors

### LINKED ISSUE #175: Install prefix paths with Windows-style backslash path separators cause several issues
Windows style paths with backslashes cause several issues -- in the worst case, a JSON decode error exception. In short, backslashes in strings get treated as escapes and gradually disappear as the install prefix gets used -- so if the install prefix contains a backslash, replace it with two backslashes so the resulting path will be `\\` instead of a single `\` that escapes the next character in some invalid way.

* `"installPrefix": "C:\abc"` -> results in JSON decode error exception
* `"installPrefix": "C:\\abc"` -> gives install paths such as `C:\u0007bc/lib64/libsomething.so` that incorrectly have the letter `a` escaped
* `"installPrefix": "C:\\\\abc"` -> gives install paths such as `C:\\abc/lib64/libsomething.so` which is correct... but not nice for the input config file installPrefix

nightlark:
Windows style paths with backslashes cause several issues -- in the worst case, a JSON decode error exception. This PR catches JSON decoding exceptions when reading from the config file and prints it with a slightly more user friendly message before exiting.

Other than that, backslashes get treated as escapes and gradually disappear as the install prefix gets used -- so if the install prefix contains a backslash, replace it with two backslashes so the resulting path will be `\\` instead of a single `\` that escapes the next character in some invalid way.

* `"installPrefix": "C:\abc"` -> results in JSON decode error exception
* `"installPrefix": "C:\\abc"` -> gives install paths such as `C:\u0007bc/lib64/libsomething.so` that incorrectly have the letter `a` escaped
* `"installPrefix": "C:\\\\abc"` -> gives install paths such as `C:\\abc/lib64/libsomething.so` which is correct... but not nice for the input config file installPrefix

Fixes #175

### DIFF
--- a/surfactant/cmd/generate.py
+++ b/surfactant/cmd/generate.py
@@ -229,7 +229,11 @@ def sbom(
 
     if pathlib.Path(config_file).is_file():
         with click.open_file(config_file) as f:
-            config = json.load(f)
+            try:
+                config = json.load(f)
+            except json.decoder.JSONDecodeError as err:
+                logger.exception(f"Invalid JSON in given config file ({config_file})")
+                raise SystemExit(f"Invalid JSON in given config file ({config_file})") from err
             # TODO: what if it isn't a JSON config file, but a single file to generate an SBOM for? perhaps file == "archive"?
     else:
         # Emulate a configuration file with the path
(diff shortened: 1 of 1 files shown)


nightlark:
To test this, use a Surfactant config file, and in the "installPrefix" field provide use some of the examples given above.

mcutshaw:
I'm able to replicate the problem and while the solution helps, I need a little bit more time to understand some escaping issues between Python/JSON.

mcutshaw (approved):
On second thought I think I understand this well enough. While I'm not 100% sure that we want to adjust backslashes to fix unescaped characters in JSON this still seems reasonable enough.

nightlark:
If you come up with some alternatives, we can continue refactoring this in future PRs -- it definitely feels a bit hacky, but it's not clear what the best way to stop Python from consuming '\`'s is.

PR #193, opened 2024-05-06
Cyt 814 docker scout

KendallHarterAtWork:
### Summary

If merged this pull request will add support for running Docker Scout on Docker .tar and .tar.gz files

### DIFF
--- a/surfactant/infoextractors/docker_file.py
+++ b/surfactant/infoextractors/docker_file.py
@@ -1,0 +1,38 @@
+# Copyright 2023 Lawrence Livermore National Security, LLC
+# See the top-level LICENSE file for details.
+#
+# SPDX-License-Identifier: MIT
+import subprocess
+import tempfile
+import gzip
+import json
+
+import surfactant.plugin
+from surfactant.sbomtypes import SBOM, Software
+
+
+def supports_file(filetype: str) -> bool:
+    return filetype in ("Docker archive tar", "Docker archive gzip")
+
+
+@surfactant.plugin.hookimpl
+def extract_file_info(sbom: SBOM, software: Software, filename: str, filetype: str) -> object:
+    if not supports_file(filetype):
+        return None
+    return extract_docker_info(filetype, filename)
+
+
+def extract_docker_info(filetype: str, filename: str) -> object:
+    if filetype == "Docker archive gzip":
+        with open(filename, 'rb') as gzip_in:
+            gzip_data = gzip_in.read()
+        with tempfile.NamedTemporaryFile() as gzip_out:
+            gzip_out.write(gzip.decompress(gzip_data))
+            return run_docker(gzip_out.name)
+    return run_docker(filename)
+
+# Function that extract_docker_info delegates to to actually run Docker scout
+def run_docker(filename: str) -> object:
+    result = subprocess.run(["docker", "scout", "sbom", "--format", "spdx", f"fs://{filename}"], capture_output=True)
+    spdx_out = json.loads(result.stdout)
+    return {'dockerSPDX': spdx_out}
--- a/surfactant/infoextractors/docker_file.py
+++ b/surfactant/infoextractors/docker_file.py
@@ -33,6 +33,6 @@ def extract_docker_info(filetype: str, filename: str) -> object:
 
 # Function that extract_docker_info delegates to to actually run Docker scout
 def run_docker(filename: str) -> object:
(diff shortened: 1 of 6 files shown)


surfactant/infoextractors/docker_file.py
```diff
@@ -0,0 +1,43 @@
...
+import subprocess
+import tempfile
+
+import surfactant.plugin
+from surfactant.sbomtypes import SBOM, Software
+
+
+def supports_file(filetype: str) -> bool:
+    return filetype in ("Docker archive tar", "Docker archive gzip")
+
```
nightlark, line 16:
```suggestion
def is_docker_scout_installed():
    # Check that Docker Scout can be run unless it's disabled
    result = subprocess.run(["docker", "scout", "--help"], capture_output=True, check=False)
    if result.returncode != 0:
        logger.warning(
            "Install Docker Scout to scan containers for additional information"
        )

# Check if Docker Scout is installed when this Python module gets loaded
disable_docker_scout = not is_docker_scout_installed()
```

The above should result in the check for Docker Scout happening just once when this module gets loaded. Another idea that came to mind is adding a "check_plugin_prereqs" hookspec that would let plugins that have choose to implement it do some checks when they first get loaded. That might be a more elegant solution, coupled with a future user settings for plugins feature.

I think in terms of UX the lack of Docker Scout shouldn't be a hard failure that requires disabling -- there are likely to be plenty of samples people analyze that don't have any Docker images. Ideally we'd know ahead of time that one of the files they'll be analyzing is a Docker image, and then warn them... a more realistic alternative to omniscience could be printing the warning in extract_docker_info() if/when a Docker file is encountered.

nightlark (commented):
Let's rename `docker_file.py` to `docker_image.py` to clarify that it is getting info from a docker image rather than a Dockerfile -- would be neat to get the original Dockerfile for an image, that would be like getting access to source code.

surfactant/infoextractors/docker_file.py
```diff
@@ -0,0 +1,43 @@
...
+        with tempfile.NamedTemporaryFile() as gzip_out:
+            gzip_out.write(gzip.decompress(gzip_data))
+            return run_docker(gzip_out.name)
+    return run_docker(filename)
+
+
+# Function that extract_docker_info delegates to to actually run Docker scout
+def run_docker(filename: str) -> object:
+    result = subprocess.run(
+        ["docker", "scout", "sbom", "--format", "spdx", f"fs://{filename}"],
```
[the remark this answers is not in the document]
KendallHarterAtWork: `docker scout cves` requires being logged into Docker's website and may also require uploading the docker image (the documentation isn't very clear on this - it kind of just assumes you're using remote images)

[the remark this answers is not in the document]
nightlark: Looking at the [docs](https://docs.docker.com/reference/cli/docker/scout/cves/), it seems to have it run on something not uploaded to DockerHub requires adding a prefix -- `archive://` would probably be it for the docker tar files we are looking at.

If it gets run on a local file, does it still require being logged into Docker? Maybe this is something that would need to be a user configurable setting that defaults to off -- not really sure why it would need logging in though, other than if Docker is maintaining their own CVE/advisory database instead of using the public databases.

surfactant/infoextractors/docker_image.py
```diff
@@ -0,0 +1,61 @@
...
+
+from loguru import logger
+
+import surfactant.plugin
+from surfactant.sbomtypes import SBOM, Software
+
+
+def is_docker_scout_installed():
+    # Check that Docker Scout can be run
+    result = subprocess.run(["docker", "scout", "--help"], capture_output=True, check=False)
```
nightlark, line 18: From the CI tests, it looks like `subprocess.run` can sometimes throw a `FileNotFoundException` (are there any others?) if the `docker` command isn't installed on a system.

surfactant/infoextractors/docker_image.py
```diff
@@ -0,0 +1,64 @@
...
+from loguru import logger
+
+import surfactant.plugin
+from surfactant.sbomtypes import SBOM, Software
+
+
+def is_docker_scout_installed():
+    # Check that Docker Scout can be run
+    try:
+        result = subprocess.run(["docker", "scout", "--help"], capture_output=True, check=False)
```
nightlark, line 19:
```suggestion
        result = subprocess.run(["docker", "scout"], capture_output=True, check=False)
```

nightlark (approved):

PR #201, opened 2024-05-15
Add options to merge command for controlling creation of system entry and type of relationships

### LINKED ISSUE #115: Add flag to disable outputting a `system` object and only create relationships
Another (external) SBOM may already define the system object -- if so, only create the relationships.

### LINKED ISSUE #116: Add an option for customizing the relationship type created by the merge command to the top-level object
Currently it is `Includes`, but there should be an option for the user to change it (with a default of `Contains`).

nightlark:
Adds several options to the merge subcommand for controlling if a system entry should be created, and what type should be used for the relationships created.

This enables workflow where relationships to a UUID should be created, but the system UUID was added to e.g. an SBOM repository as part of a separate file.

The more "standard" `Contains` relationship is used instead of the `Includes` relationship that apparently was unique to Surfactant merged SBOMs.

## Changes
* Add option for disabling the creation of a system object
* Add option for specifying what type of relationship should be created for relationships to the system object; use "Contains" as the default type
* Fix a bug where a System SBOM object accessed UUID using subscript notation (`["UUID"]`) instead of as a member variable (`.UUID`)

Resolves #115
Resolves #116

### DIFF
--- a/surfactant/cmd/merge.py
+++ b/surfactant/cmd/merge.py
@@ -1,11 +1,12 @@
 import json
 import uuid as uuid_module
 from collections import deque
-from typing import Dict, List
+from typing import Dict, List, Tuple
 
 import click
 from loguru import logger
 
+from surfactant.configmanager import ConfigManager
 from surfactant.plugin.manager import find_io_plugin, get_plugin_manager
 from surfactant.sbomtypes._relationship import Relationship
 from surfactant.sbomtypes._sbom import SBOM
@@ -14,11 +15,18 @@ from surfactant.sbomtypes._system import System
 
 @click.argument("sbom_outfile", envvar="SBOM_OUTPUT", type=click.File("w"), required=True)
 @click.argument("input_sboms", type=click.File("r"), required=True, nargs=-1)
-@click.option("--config_file", type=click.File("r"), required=False)
+@click.option(
+    "--config_file",
+    type=click.File("r"),
+    required=False,
+    help="Config file for controlling some aspects of the merged SBOM, primarily the creation of a new top-level system object (settings here will typically take precedence over command line options)",
+)
 @click.option(
     "--output_format",
     is_flag=False,
-    default="surfactant.output.cytrics_writer",
+    default=ConfigManager().get(
+        "core", "output_format", fallback="surfactant.output.cytrics_writer"
+    ),
     help="SBOM output format, options=surfactant.output.[cytrics|csv|spdx]_writer",
 )
 @click.option(
@@ -27,8 +35,36 @@ from surfactant.sbomtypes._system import System
     default="surfactant.input_readers.cytrics_reader",
     help="SBOM input format, assumes that all input SBOMs being merged have the same format, options=surfactant.input_readers.[cytrics|cyclonedx|spdx]_reader",
 )
+@click.option(
+    "--system_uuid",
+    is_flag=False,
+    help="System UUID to use for relationships to a top-level system object",
+)
+@click.option(
+    "--system_relationship",
+    is_flag=False,
+    default="Contains",
+    show_default=True,
+    help="Relationship type between merged SBOM contents to a top-level system object",
+)
+@click.option(
+    "--add_system/--no_add_system",
+    default=False,
+    show_default=True,
+    help="Create a top-level system entry for tying together the merged SBOM components. When disabled, relationships will still be created to a provided system UUID",
+)
 @click.command("merge")
-def merge_command(input_sboms, sbom_outfile, config_file, output_format, input_format):
+# pylint: disable-next=too-many-positional-arguments
+def merge_command(
+    input_sboms,
+    sbom_outfile,
+    config_file,
+    output_format,
+    input_format,
+    system_uuid,
+    system_relationship,
+    add_system,
+):
     """Merge two or more INPUT_SBOMS together into SBOM_OUTFILE.
 
     An optional CONFIG_FILE can be supplied to specify a root system entry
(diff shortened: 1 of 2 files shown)


surfactant/cmd/merge.py
```diff
@@ -26,8 +31,29 @@
...
+)
+@click.option(
+    "--add_system/--no_add_system",
+    default=False,
+    show_default=True,
+    help="Create a top-level system entry for tying together the merged SBOM components. When disabled, relationships will still be created to a provided system UUID",
+)
 @click.command("merge")
-def merge_command(input_sboms, sbom_outfile, config_file, output_format, input_format):
+def merge_command(
```
shaynakapadia, line 48: If you want to clean this up a bit, you could use **kwargs. But it isn't neccessary

shaynakapadia:
@nightlark I'm a bit concerned with the functionality of creating a contains relationship with a randomly generated uuid if no system uuid is present and the user specifies they don't want one added. This would create a relationship to a uuid that has no information attached to it, so a user looking at it wouldn't know if it was a system uuid, software uuid, or what. I think it would be better to require the user to provide a uuid to link to if no system uuid is present and a system object is not to be added. This would prevent floating UUIDs with no meaning attached to them. Thoughts?

shaynakapadia:
@nightlark Also I typically use Includes for merges, what was the reason for changing the default to Contains?

shaynakapadia:
If possible I would also add a few unit tests!

surfactant/cmd/merge.py
nightlark: I've been thinking about how to handle this pattern in click that seems to result in function definitions with a lot of arguments -- ideally I'd like to have a function that makes the feature usable when using Surfactant as a library, which will probably need at least one function with many args.

**kwargs could be a nice solution to avoid having two functions with duplicate function signatures that have many args and need to be kept in sync. I'll look into this some more and probably convert the generate subcommand at the same time.

nightlark:
> @nightlark Also I typically use Includes for merges, what was the reason for changing the default to Contains?

Mostly consistency -- I was talking with Aaron about what relationships other labs were using in their SBOMs for tying software entries to the system object, and it seems everyone else is using "Contains".

> @nightlark I'm a bit concerned with the functionality of creating a contains relationship with a randomly generated uuid if no system uuid is present and the user specifies they don't want one added. This would create a relationship to a uuid that has no information attached to it, so a user looking at it wouldn't know if it was a system uuid, software uuid, or what. I think it would be better to require the user to provide a uuid to link to if no system uuid is present and a system object is not to be added. This would prevent floating UUIDs with no meaning attached to them. Thoughts?

Oof... that would be fun to figure out why an SBOM fails validation when uploading to the repository. I'll add a command line option for providing a system_uuid, and checks to make sure a relationship to a random UUID is only added if the system was also added. I'll also make so that command line options take precedence over the config file like I'd originally planned.

surfactant/cmd/merge.py
shaynakapadia: Ah I see, I think you can achieve this by creating a custom decorator that adds options or args in the way you want. Similar to what you see in this stack overflow post: https://stackoverflow.com/questions/50061342/is-it-possible-to-reuse-python-click-option-decorators-for-multiple-commands

shaynakapadia:
> Mostly consistency -- I was talking with Aaron about what relationships other labs were using in their SBOMs for tying software entries to the system object, and it seems everyone else is using "Contains".
> 

Ah got it, maybe I'm using them wrong then :D

surfactant/cmd/merge.py
```diff
@@ -42,10 +79,19 @@ def merge_command(input_sboms, sbom_outfile, config_file, output_format, input_f
...
+
+
+# pylint: disable-next=too-many-positional-arguments
+def merge(
+    input_sboms,
+    sbom_outfile,
+    config,
+    output_writer,
+    system_uuid=None,
+    system_relationship="Contains",
```
shaynakapadia, line 92: default is already specified in the click option, do we need to specify it here too?

surfactant/cmd/merge.py
```diff
@@ -42,10 +79,19 @@ def merge_command(input_sboms, sbom_outfile, config_file, output_format, input_f
...
+    add_system=True,
```
shaynakapadia, line 93: default here is True, but click default is False

shaynakapadia:
@nightlark The fix for the random uuid looks good! The logic if random system uuid is used, you must add the system object to the sbom makes sense to me and will prevent any floating uuids from getting generated.

surfactant/cmd/merge.py
nightlark: Added a comment reminder to keep in-sync with click default -- I think the only time this default value would be used is if someone is using Surfactant as a library and directly calling this merge function from within their own Python scripts.

surfactant/cmd/merge.py
```diff
@@ -151,23 +202,34 @@ def dfs(rel):
...
+    if system_uuid:
+        system["UUID"] = system_uuid
+    elif "UUID" not in system:
+        # No UUID, generate a random one...
+        using_random_uuid = True
         system["UUID"] = str(uuid_module.uuid4())
+    # check if the UUID appears valid based on the CyTRICS schema
+    elif not sbom.is_valid_uuid4(system["UUID"]):
+        invalid_uuid = system["UUID"]
+        logger.error(f"Invalid uuid4 given ({invalid_uuid}) for the system")
```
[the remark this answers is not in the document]
nightlark: UUID4 is a standard format -- don't need to add extra verbosity to it.

surfactant/cmd/merge.py
```diff
@@ -56,22 +101,28 @@ def merge(input_sboms, sbom_outfile, config, output_writer):
...
     # Add a system relationship to each root software/systems entry identified
-    for r in roots:
-        merged_sbom.relationships.add(
-            Relationship(xUUID=system["UUID"], yUUID=r, relationship="Includes")
+    if not using_random_uuid or add_system:
+        if config and "systemRelationship" in config:
+            system_relationship = config["systemRelationship"]
+        for r in roots:
+            merged_sbom.relationships.add(
+                Relationship(xUUID=system.UUID, yUUID=r, relationship=system_relationship)
```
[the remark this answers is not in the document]
nightlark: system_relationship is determined earlier in the function.

PR #209, opened 2024-05-23
Implement CLI add subcommand

shaynakapadia:
### Summary

This merge request implements the basic functionality of the surfactant cli add command. 

If merged this pull request will allow users to add details to sboms without having to do so manually. They can add installPaths or containerPaths, add entries directly or by file, and add relationships. In combination with the find command, they can edit subsets of an sbom without using a IDE. 

### Proposed changes

The proposed changes are
- Adding the cli_add class which handles the details of how items are added to an sbom
- Creating the add command function which takes in the command line inputs and sends them to the cli_add class

### Usage
Some examples of usage are listed below

Adds a relationship to the "relationships" key in sbom.json
```bash
surfactant cli add --relationship "{xUUID:"123",yUUID:456, "relationship: "Uses"}" sbom.json
```
Adds a direct entry into the "software" key in sbom.json
```bash
surfactant cli add --entry "{UUID:"123",filename:"test.exe", "sha256": "3423csdlkf13048kj"}" sbom.json
```
Creates an entry for test.exe in the "software" key in sbom.json
```bash
surfactant cli add --file test.exe sbom.json
```
Redirects output to a new file called subset.json
```bash
surfactant cli add --file test.exe --output subset.json sbom.json
```
### Example Workflow with cli find
Say I wanted to add an installPath to a subset of the entries in the sample sbom helics_sbom.json. I could do so with the following steps
```bash
surfactant cli find --uuid 477da45b-bb38-450e-93f7-e525aaaa6862 helics_sbom.json > subset.json
surfactant cli add --installPath "C:\Users\Downloads" subset.json 
surfactant merge helics_sbom.json subset.json final.json
```

Open to suggestions on the workflow as well. 

Resolves CYT-690

### DIFF
--- a/surfactant/cmd/cli.py
+++ b/surfactant/cmd/cli.py
@@ -175,7 +175,7 @@ class cli_add:
         self.sbom.software.append(Software.from_dict(entry))
 
     def add_installpath(self, prefixes: tuple):
-        cleaned_prefixes = tuple([p.strip("/") for p in prefixes])
+        cleaned_prefixes = tuple([p.rstrip("/") for p in prefixes])
         containerPathPrefix, installPathPrefix = cleaned_prefixes
         for sw in self.sbom.software:
             for path in sw.containerPath:
--- a/surfactant/cmd/cli.py
+++ b/surfactant/cmd/cli.py
@@ -84,7 +84,7 @@ def find(sbom, output_format, input_format, **kwargs):
     "--input_format",
     is_flag=False,
     default="surfactant.input_readers.cytrics_reader",
-    help="SBOM input format, assumes that all input SBOMs being merged have the same format, options=[cytrics|cyclonedx|spdx]",
+    help="SBOM input format, options=[cytrics|cyclonedx|spdx]",
 )
 @click.command("add")
 def add(sbom, output, output_format, input_format, **kwargs):
@@ -117,11 +117,12 @@ def edit(sbom, output_format, input_format, **kwargs):
 
 class cli_add:
     """
-    A class that implements the surfactant cli find functionality
+    A class that implements the surfactant cli add functionality
 
     Attributes:
-    match_functions     A dictionary of functions that provide matching functionality for given SBOM fields (i.e. uuid, sha256, installpath, etc)
-    sbom                An internal record of sbom entries the class adds to as it finds more matches.
+    match_functions         A dictionary of functions that provide matching functionality for given SBOM fields (i.e. uuid, sha256, installpath, etc)
+    camel_case_conversions  A dictionary of string conversions from all lowercase to camelcase. Used to convert python click options to match the SBOM attribute's case
+    sbom                    An internal record of sbom entries the class adds to as it finds more matches.
     """
 
     camel_case_conversions: dict
@@ -155,7 +156,7 @@ class cli_add:
 
     def execute(self, input_sbom: SBOM, **kwargs):
         """Executes the main functionality of the cli_find class
-        param: input_sbom   The sbom to find matches within
+        param: input_sbom   The sbom to add entries to
         param: kwargs:      Dictionary of key/value pairs indicating what features to match on
         """
         converted_kwargs = self.handle_kwargs(kwargs)
@@ -191,8 +192,9 @@ class cli_find:
     A class that implements the surfactant cli find functionality
 
     Attributes:
-    match_functions     A dictionary of functions that provide matching functionality for given SBOM fields (i.e. uuid, sha256, installpath, etc)
-    sbom                An internal record of sbom entries the class adds to as it finds more matches.
+    match_functions         A dictionary of functions that provide matching functionality for given SBOM fields (i.e. uuid, sha256, installpath, etc)
+    camel_case_conversions  A dictionary of string conversions from all lowercase to camelcase. Used to convert python click options to match the SBOM attribute's case
+    sbom                    An internal record of sbom entries the class adds to as it finds more matches.
     """
 
     match_functions: dict
--- a/surfactant/cmd/cli.py
+++ b/surfactant/cmd/cli.py
@@ -66,13 +66,8 @@ def find(sbom, output_format, input_format, **kwargs):
     "--installPath",
     is_flag=False,
     type=str,
-    help="Adds installPath to add to all entries",
-)
-@click.option(
-    "--containerPath",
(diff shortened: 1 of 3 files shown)


surfactant/cmd/cli.py
```diff
@@ -52,16 +55,135 @@ def find(sbom, output_format, input_format, **kwargs):
...
+    def add_entry(self, entry):
+        self.sbom.software.append(Software.from_dict(entry))
+
+    def add_installpath(self, path):
+        for sw in self.sbom.software:
+            sw.installPath.append(path)
+
+    def add_containerpath(self, path):
+        for sw in self.sbom.software:
+            sw.containerPath.append(path)
```
nightlark, line 186: I don't think these will work as expecte  since the CyTRICS SBOM schema uses paths that includes the file name at the end (e.g. `/usr/local/bin/mybinary`, which helps distinguish the install/container paths for multiple files that have different names installed to different locations (or symlinks).

My naive take would be that the provided install/container path should be a "prefix" that gets prepended to every file name associated with each software entry, but that would result in paths that just look like every single file is in the same directory. The problem I see is that the SBOM schema(s) don't preserve enough information on what subdirectory each file should be in relative to an install/container path prefix (e.g. the middle part of `<install/container_prefix><subdirectories><filename>` ).

Maybe this functionality should require supplying a "original prefix", then any time a container/install path has that prefix, add a new entry that has it swapped out for a new prefix?

shaynakapadia: Ah I see that is a good point. I think your suggestion is a good option for a first pass, but I do think it is easier to just find and replace within a code editor than to do it via command line. Would there be a benefit to having this functionality as a cli option? I guess if someone wanted to script edits using surfactant, I can see them using this instead of VSCode or another IDE

surfactant/cmd/cli.py
```diff
@@ -52,16 +55,135 @@ def find(sbom, output_format, input_format, **kwargs):
...
 
-@click.argument("sbom", type=click.File("r"), required=True)
-@click.command("add")
-def add(sbom):
-    "CLI command to add specific entry(s) to a supplied SBOM"
+class cli_add:
+    """
+    A class that implements the surfactant cli find functionality
+
+    Attributes:
```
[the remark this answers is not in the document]
shaynakapadia: Oops, forgot to add that description: Its a dictionary of string conversions from all lowercase to camelcase. Used to convert python click options to match the SBOM attribute's case

surfactant/cmd/cli.py
nightlark: Args to cli command: abc0/ -> /usr/local
containerpath (abc0/bin/helics_binary) -> installpath (/usr/local/bin/helics_binary)

tests/cmd/test_cli.py
```diff
@@ -91,3 +91,59 @@ def test_find_with_bad_filter():
...
+    previous_software_len = len(in_sbom.software)
+    out_bom = cli_add().execute(
+        in_sbom, file=pathlib.Path(__file__).parent / "../data/a_out_files/big_m68020.aout"
+    )
+    assert len(out_bom.software) == previous_software_len + 1
+    assert (
+        out_bom.software[8].sha256
+        == "9e125f97e5f180717096c57fa2fdf06e71cea3e48bc33392318643306b113da4"
+    )
+    del in_sbom.software[-1]
```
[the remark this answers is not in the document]
shaynakapadia: Good idea! Didn't even know that was a thing. Updated tests to use the "test_sbom" test fixture.

surfactant/cmd/cli.py
```diff
@@ -52,25 +55,141 @@ def find(sbom, output_format, input_format, **kwargs):
...
+        self.sbom.add_relationship(Relationship(**value))
+
+    def add_file(self, path):
+        self.sbom.software.append(Software.create_software_from_file(path))
+
+    def add_entry(self, entry):
+        self.sbom.software.append(Software.from_dict(entry))
+
+    def add_installpath(self, prefixes: tuple):
+        containerPathPrefix, installPathPrefix = prefixes
```
nightlark, line 178: I think trailing `/` (or `\` though even on Windows users should favor `/`) should be trimmed from the containerPathPrefix and installPathPrefix, and then as part of the append we can manually add in the path separator if needed. Actually, if the existing container path in the SBOM is correct and the user didn't mess up and type a partial directory, the existing path separator that is in there should just work without us adding anything.

The cases that are currently odd for container path and install path prefix pairs, for manipulating a container path of `abc/file123`:
* abc/ /local -> /localfile123
* abc /local/ -> /local//file123

shaynakapadia: I added the removal of trailing `/` 
So with your examples, the result should be: 
Original containerpath: abc/file123
abc/ /local -> stripped to abc /local -> /local/file123
abc /local/ -> stripped to abc /local -> /local/file123
/abc /local -> nothing to strip -> no match since /abc != abc
Does that sound right?

nightlark: Yea, that sounds right.

surfactant/cmd/cli.py
```diff
@@ -52,25 +55,142 @@ def find(sbom, output_format, input_format, **kwargs):
...
+
+    def add_file(self, path):
+        self.sbom.software.append(Software.create_software_from_file(path))
+
+    def add_entry(self, entry):
+        self.sbom.software.append(Software.from_dict(entry))
+
+    def add_installpath(self, prefixes: tuple):
+        cleaned_prefixes = tuple([p.rstrip("/") for p in prefixes])
+        containerPathPrefix, installPathPrefix = cleaned_prefixes
```
nightlark, line 179: pre-commit thinks a generator could be used here... if we stick with a tuple, this could be changed to `cleaned_prefixes = tuple(cleaned_prefixes = tuple(p.rstrip("/") for p in prefixes)` to get rid of the list comprehension that it is probably complaining about.

The generator approach would skip the packing into a tuple then unpacking and end up as something like:
```python
cleaned_prefixes = (p.rstrip("/") for p in prefixes)
containerPathPrefix, installPathPrefix = cleaned_prefixes
```

shaynakapadia: Ah I keep forgetting to check the pre-commit hooks. Ive updated with your generator suggestion

nightlark (approved):

PR #216, opened 2024-06-14
Fix: plugins running multiple times on the same file due to symlinks

### LINKED ISSUE #215: Extract info hooks run multiple times on files
File info extraction hooks run multiple times on a file, which causes problems for plugins that write extra output files having duplicate info.

The underlying issue is around here: https://github.com/LLNL/Surfactant/blob/main/surfactant/cmd/generate.py#L367
From what I recall, `get_software_entry` is called to get the hash of the underlying file for tracking what symlinks point to that file -- which also gives plugins an opportunity to run. This should be changed so that if the file is a symlink it computes the file hash without trying to reuse `get_software_entry`.

This should also simplify the control flow some by not needing the `file_is_symlink` flag.

nightlark:
### Summary

If merged this pull request will refactor the logic used for handling symlinks so that plugins do not run multiple times for the same file, along with changing the way symlink file names are added so they don't depend on the user specifying an install prefix.

Plugins will still run multiple times if there are separate files that happen to have identical hashes -- which could cause a similar problem with the cve-bin-tool plugin as this change solves for symlinks. Some plugins might need to gather information from other files relative to the one they are looking at.
1. Should an extra optional argument be added to the info extractor hook to tell the plugin if a file with the same hash has previously been seen, so it can adjust its behavior accordingly?
2. Should a separate info extraction hook be added for duplicate files in different locations?

### Proposed changes

- Refactor symlink handling to prevent plugins from running multiple time on the same file due to symlinks
- Add file name symlinks (+ track them in the metadata section) regardless of the user supplied an install prefix

Resolves #215

### DIFF
--- a/surfactant/cmd/generate.py
+++ b/surfactant/cmd/generate.py
@@ -13,6 +13,7 @@ import click
 from loguru import logger
 
 from surfactant import ContextEntry
+from surfactant.fileinfo import sha256sum
 from surfactant.plugin.manager import find_io_plugin, get_plugin_manager
 from surfactant.relationships import parse_relationships
 from surfactant.sbomtypes import SBOM, Software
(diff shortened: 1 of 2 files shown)


surfactant/cmd/generate.py
```diff
@@ -380,16 +406,10 @@ def sbom(
...
-                                # Track the symlink, but don't add to list of entries
-                                # as it'll be processed later anyways
-                                if sw_parent.sha256 not in file_symlinks:
-                                    file_symlinks[sw_parent.sha256] = []
-                                file_symlinks[sw_parent.sha256].append(install_filepath)
-                            else:
-                                entries.append(sw_parent)
-                                for sw in sw_children:
-                                    entries.append(sw)
+                            print(f"===Adding {sw_parent.installPath}")
```
KendallHarterAtWork, line 409: This print statement should use the logger instead

nightlark: Oh, I forgot to remove that print.

KendallHarterAtWork (approved):

PR #217, opened 2024-06-14
Recommend pipx for user installation and show how to pip install plugins

### LINKED ISSUE #204: Update install directions: pipx/uv
Update the install directions to recommend pipx for regular users who aren't doing development (avoids having to manage venv and future built-in Surfactant command may have some ability to update plugins in its own venv). Mention pipx inject for managing plugins (runs venv process behind the scenes).

Also suggest uv tool add functionality as an alternative.

nightlark:
### Summary

If merged this pull request will update the docs to recommend users install Surfactant with pipx (which hides the manual virtual environment management) and adds an example of installing plugins.

### Proposed changes

- Recommend pipx for user installation
- Add some more detail on how to install plugins

Resolves #204

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -43,6 +43,8 @@ To install optional dependencies required for running pytest and pre-commit:
 pip install -e ".[test,dev]"
 ```
 
+`pip install` with the `-e` or `--editable` option can also be used to install Surfactant plugins for development.
+
 ## Code of Conduct
 
 All participants in the Surfactant community are expected to follow our [Code of Conduct](https://www.contributor-covenant.org/version/2/1/code_of_conduct.html).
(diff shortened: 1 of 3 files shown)


shaynakapadia (approved):
Looks good! Tested the steps out and they make sense to me.

mcutshaw:
@nightlark So to confirm for editable installs we're still recommending `pip install -e`? It looks like `pipx install --force --editable .` will also create an editable install.

mcutshaw (approved):

nightlark:
For now we're still recommending pip for an editable install -- I haven't played with using editable pipx installs for development yet, so I'm not sure how that could make things more difficult (e.g. installing a plugin within the editable install, multiple venvs for different feature branches/Python versions, etc).

PR #221, opened 2024-06-18
Add CycloneDX read SBOM hook implementation

mws180000:
Latest working draft of the CycloneDX input SBOM hook.

- This includes adding any vulnerabilities from the CycloneDX SBOM into the observations section of the CyTRICS SBOM

- Change in mapping requirements: It turns out that a CycloneDX component can have nested subcomponents. Mapping these subcomponents to the components section of the CyTRICS software entry format would fit better than mapping the CycloneDX services section to the CyTRICS components section since CycloneDX services are it's won section in the SBOM (rather than being part of individual CycloneDX component entries)

- Some CycloneDX-exclusive component data is currently added to the metadata section of the corresponding CyTRICS software entry, while the rest needs to be transformed into formats that are JSON-serializable


nightlark:
In the CyTRICS schema working group I was thinking that it would be nice to be able to track what CycloneDX/SPDX identifier is associated with software entries, which would enable SBOM conversion to preserve a stable identifier for a roundtrip conversion from CycloneDX -> CyTRICS/internal -> CycloneDX.

wangmot:
`cyclonedx_reader.py` script is not finished yet and is still a bit rough. Most cyclonedx fields are mapped out with comments and some example extraction code. 

Some notes:
-Ran into some serialization issues with objects
-script needs more testing with sample cyclonedx SBOMs with as many fields as possible
-code could be refactored for readability and reusability (ex. helper functions)
-for the most part, I've kind of fleshed out how to do the mappings to the surfactant sbom schema, but some fields may still need adjustment

PR #229, opened 2024-07-22
Pin dependency versions

KendallHarterAtWork:
### Summary

If merged this pull request will pin dependency versions to known compatible versions

### Proposed changes

Pin patch versions for `0.*`, pin minor versions for `>0.*`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -50,7 +50,7 @@ dependencies = [
     "cyclonedx-python-lib==7.5.1",
     "pluggy==1.*",
     "click==8.*",
-    "javatools>=1.6.0",
+    "javatools==1.*",
     "loguru==0.7.*",
     "flask==3.*"
 ]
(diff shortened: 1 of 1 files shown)


nightlark (approved):

pyproject.toml
```diff
@@ -39,19 +39,20 @@ classifiers = [
...
     "spdx-tools==0.8.*",
     "cyclonedx-python-lib==7.5.1",
-    "pluggy",
-    "click",
-    "javatools>=1.6.0",
-    "loguru",
-    "flask"
+    "pluggy==1.*",
+    "click==8.*",
+    "javatools==1.*",
```
nightlark, line 53:
```suggestion
    "javatools>=1.6,==1.*",
```

nightlark (changes requested):
Actually, iirc there was a bug that prevented versions of javatools prior to 1.6.0 from working -- so let's keep that as an additional version constraint.

nightlark (approved):

Issue #100, opened 2023-12-15
User configurable options for plugins


nightlark:
**Is your feature request related to a problem? Please describe.**
Users should be able to configure settings unique to each plugin, and ideally be able to control which plugins run beyond using pip virtual environments. This feature could also be used to let users configure default Surfactant settings, such as the recorded institution and potentially preferred behavior for various scenarios. A `config` or `settings` subcommand for editing a user-wide settings file similar to `git config` could be nice.

**Describe the solution you'd like**
From discussion in the working group, the following is what we'd like (copy+paste from PowerPoint slides gives an image):
![image](https://github.com/LLNL/Surfactant/assets/3969255/670ceefb-b941-4fc3-89b9-b9ca7d8de079)

nightlark:
For plugins developers communicating the supported options, I think a docstring will be used, with a heading similar to the one used for arguments that will list config options supported. Future functionality could include parsing the docstring for all loaded plugins and presenting a list of supported config options.

PR #231, opened 2024-07-22
Add support for user config file for plugin options

nightlark:
### Summary

If merged this pull request will add support for a user config file that can be used to control plugin settings, and a new `config` subcommand that is modeled after `git config` for getting and setting config options.

Partly addresses #100 -- does not add a way to specify settings via the command line. For plugin config options supported, I think we will have developers include a list of options in their docstrings.

### Proposed changes

- Add ConfigManager that can be used by Surfactant plugins to support user configuration options, stored in a standard location in the user's home directory
- Add a `config` subcommand based on `git config` that can be used to modify the configuration file from the command line
- Add tests for the ConfigManager class
- Add documentation on how to set/get configuration options, from both a user and developer perspective

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,18 +102,18 @@ Surfactant settings can be changed using the `surfactant config` subcommand, or 
 
 ### Command Line
 
-Using `surfactant config` is very similar to basic use of `git config`. The key whose value is being accessed will be in the form `section.option` where `section` is typically a plugin name or `core`, and `option` is the option to set. As an example, a plugin that includes an entire copy of the original files as part of the generated SBOM might provide an option to specify the format.
+Using `surfactant config` is very similar to the basic use of `git config`. The key whose value is being accessed will be in the form `section.option` where `section` is typically a plugin name or `core`, and `option` is the option to set. As an example, the `core.recorded_institution` option can be used to configure the recorded institution used to identify who the creator of a generated SBOM was.
 
-Setting the configuration option to `base64` could be done using:
+Setting this option to `LLNL` could be done with the following command:
 
 ```bash
-surfactant config include_entire_file.format base64
+surfactant config core.recorded_institution LLNL
 ```
 
 Getting the currently set value for the option would then be done with:
 
 ```bash
-surfactant config include_entire_file.format
+surfactant config core.recorded_institution
 ```
 
 ### Manual Editing
@@ -126,8 +126,8 @@ to `~/.config`. On Windows, the file is stored in the Roaming AppData folder at 
 The file itself is a TOML file, and for the previously mentioned example plugin may look something like this:
 
 ```toml
-[include_entire_file]
-format = "base64"
+[core]
+recorded_institution = "LLNL"
 ```
 
 ## Usage
--- a/README.md
+++ b/README.md
@@ -136,9 +136,9 @@ recorded_institution = "LLNL"
 
 In order to test out surfactant, you will need a sample file/folder. If you don't have one on hand, you can download and use the portable .zip file from <https://github.com/ShareX/ShareX/releases> or the Linux .tar.gz file from <https://github.com/GMLC-TDC/HELICS/releases>. Alternatively, you can pick a sample from https://lc.llnl.gov/gitlab/cir-software-assurance/unpacker-to-sbom-test-files
 
-### Build sample configuration file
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -96,15 +96,49 @@ pip install -e ".[test,dev]"
 pip install -e plugins/fuzzyhashes
 ```
 
+## Settings
+
+Surfactant settings can be changed using the `surfactant config` subcommand, or by hand editing the settings configuration file (this is not the same as the JSON file used to configure settings for a particular sample that is described later).
+
+### Command Line
+
+Using `surfactant config` is very similar to basic use of `git config`. The key whose value is being accessed will be in the form `section.option` where `section` is typically a plugin name or `core`, and `option` is the option to set. As an example, a plugin that includes an entire copy of the original files as part of the generated SBOM might provide an option to specify the format.
```
shaynakapadia, line 105:
```suggestion
Using `surfactant config` is very similar to the basic use of `git config`. The key whose value is being accessed will be in the form `section.option` where `section` is typically a plugin name or `core`, and `option` is the option to set. As an example, a plugin that includes an entire copy of the original files as part of the generated SBOM might provide an option to specify the format.
```

README.md
```diff
@@ -96,15 +96,49 @@ pip install -e ".[test,dev]"
...
+Setting the configuration option to `base64` could be done using:
+
+```bash
+surfactant config include_entire_file.format base64
+```
+
+Getting the currently set value for the option would then be done with:
+
+```bash
+surfactant config include_entire_file.format
```
shaynakapadia, line 116: Is this a working example or does this setting not exist?

README.md
```diff
@@ -96,15 +96,49 @@ pip install -e ".[test,dev]"
...
+```
+
 ## Usage
 
 ### Identify sample file
 
 In order to test out surfactant, you will need a sample file/folder. If you don't have one on hand, you can download and use the portable .zip file from <https://github.com/ShareX/ShareX/releases> or the Linux .tar.gz file from <https://github.com/GMLC-TDC/HELICS/releases>. Alternatively, you can pick a sample from https://lc.llnl.gov/gitlab/cir-software-assurance/unpacker-to-sbom-test-files
 
-### Build configuration file
+### Build sample configuration file
```
shaynakapadia, line 139:
```suggestion
### Build configuration file for sample
```
Not sure if this change is better, but my immiedate thought was to build a sample configuration file (like an example one rather than one for a software sample). We could change this phrasing to make it a bit more clear

README.md
```diff
@@ -96,15 +96,49 @@ pip install -e ".[test,dev]"
...
 
-A configuration file contains the information about the sample to gather information from. Example JSON configuration files can be found in the examples folder of this repository.
+A sample configuration file contains the information about the sample to gather information from. Example JSON sample configuration files can be found in the examples folder of this repository.
```
shaynakapadia, line 141:
```suggestion
A configuration file for a sample contains the information about the sample to gather information from. Example JSON sample configuration files can be found in the examples folder of this repository.
```

README.md
nightlark: The setting doesn't exist, so it won't actually affect how Surfactant runs. It works in the sense that setting and getting the value will save the value to the underlying config file and retrieve the stored value.

README.md
shaynakapadia: It becomes more clear when you read the rest of the documentation, so this change might not be neccessary

README.md
nightlark: That makes sense. The issue I wanted to address is how do we refer to different types of configuration/setting files, to avoid confusing when talking about them? Is there a prefix or alternate name we could use instead? e.g. "SBOM configuration file", "project file", etc

README.md
nightlark: I might update this to use setting the recorded institution as an example (along with making that one of the first config options available).

nightlark: This has been updated to be the recorded_institution setting now.

README.md
shaynakapadia: Not sure which one is best but here are some more suggestions:
"Specimen Configuration File", "Software Sample Configuration File", or "Input Configuration File"

shaynakapadia (approved):
Looks good!

Issue #126, opened 2024-02-06
UX: Generate SBOM info for single file


nightlark:
If the generate command gets given a binary file, an empty SBOM may be generated. It could make sense to detect that it is a non-JSON/config file, and generate a single entry SBOM (similar to giving Surfactant a directory). Same for listing a file under extract paths in a config file instead of a folder.

mcutshaw:
So the only thing that comes to mind for checking if the first file is non-JSON/config file is trying to open it as JSON and if it fails assume its a target to analyze. If that's acceptable I can work on a pull for this one.

nightlark:
That's probably a decent option... maybe paired with checking to see if the file name ends in .json?

gliese1337:
I tried creating a config file with a single executable file name in the "extractPaths", and it still generated an empty SBOM. Copying the executable to a new empty directory and then pointing Surfactant at the directory works, but that's very unergonomic. It would be nice to have single-file support at least in the config, if not as a simple command-line option.

nightlark:
I'd actually just been thinking about that with things listed in extractPaths a week or two back -- adding single file support to that should be relatively simple compared to as part of the (currently overloaded) positional argument in the command line.

I'll see try to get that change made soon.

nightlark:
For command line arguments, something I've seen a number of tools do is support some special prefixes at the start of the provided argument that specify exactly what the argument is. For example, a tool (Syft) that handles a variety of container images uses:

```bash
docker-archive:path/to/yourimage.tar     use a tarball from disk for archives created from "docker save"
oci-archive:path/to/yourimage.tar        use a tarball from disk for OCI archives (from Skopeo or otherwise)
oci-dir:path/to/yourimage                read directly from a path on disk for OCI layout directories (from Skopeo or otherwise)
singularity:path/to/yourimage.sif        read directly from a Singularity Image Format (SIF) container on disk
dir:path/to/yourproject                  read directly from a path on disk (any directory)
file:path/to/yourproject/file            read directly from a path on disk (any single file)
```

I could imagine doing something like having special `dir`, `file`, and `config` prefixes that specify exactly what the file is, possibly combined with a heuristic like if the file provided is a json file, then assume it is a config file by default.

PR #244, opened 2024-08-21
Add support for extract paths being files

nightlark:
### Summary

If merged this pull request will allow extract paths given in a specimen config file to be a file.

### Proposed changes

- Allow extract paths to be individual files (os.walk ignores it, so it had to be a separate case)
- Refactor some of the code for adding entries and determining the install prefix into separate functions to reduce duplicate code
- Add code to handle some corner cases involving trailing slashes that could result in messed up container paths and install paths

Part of addressing #126.


nightlark:
@gliese1337 here's a PR that adds support for single files in the extract paths for the specimen config files -- feel free to leave a review with any issues you find.

Issue #238, opened 2024-08-05
Add config option for including all files in SBOM


nightlark:
Add a configuration option for including all files in an SBOM, regardless of if they are a recognized type or not.

nightlark:
This should also be added as an option to the specimen config files, so that infoextractors that extract files (or a file) from other files can say that the extract files should all be included in the SBOM. This could allow forcing inclusion of the binary file extracted from a s-rec or hex file, even if that file has no magic bytes identifying it as a file type Surfactant recognizes.

PR #246, opened 2024-08-27
Add option to include all files

KendallHarterAtWork:
### Summary

If merged this pull request will add an option to include all files found in the SBOM, rather than just those identified by Surfactant.  This addresses #238

### Proposed changes

There are two ways to include all files, either passing the `--include_all_files` option to the generate subcommand, or adding and setting to true an `includeAllFiles` option in the input file when specifying extract paths and install paths.

### DIFF
--- a/docs/basic_usage.md
+++ b/docs/basic_usage.md
@@ -23,7 +23,7 @@ $  surfactant generate [OPTIONS] CONFIG_FILE SBOM_OUTFILE [INPUT_SBOM]
 **--output_format**: (optional) changes the output format for the SBOM (given as full module name of a surfactant plugin implementing the `write_sbom` hook)\
 **--input_format**: (optional) specifies the format of the input SBOM if one is being used (default: cytrics) (given as full module name of a surfactant plugin implementing the `read_sbom` hook)\
 **--help**: (optional) show the help message and exit\
-**--inclulde_all_files**: Include all files in the SBOM, rather than just those recognized by Surfactant
+**--include_all_files**: (optional) include all files in the SBOM, rather than just those recognized by Surfactant
 
 
 ## Merging SBOMs
--- a/docs/basic_usage.md
+++ b/docs/basic_usage.md
@@ -22,7 +22,8 @@ $  surfactant generate [OPTIONS] CONFIG_FILE SBOM_OUTFILE [INPUT_SBOM]
 **--recorded_institution**: (optional) the name of the institution collecting the SBOM data (default: LLNL)\
 **--output_format**: (optional) changes the output format for the SBOM (given as full module name of a surfactant plugin implementing the `write_sbom` hook)\
 **--input_format**: (optional) specifies the format of the input SBOM if one is being used (default: cytrics) (given as full module name of a surfactant plugin implementing the `read_sbom` hook)\
-**--help**: (optional) show the help message and exit
+**--help**: (optional) show the help message and exit\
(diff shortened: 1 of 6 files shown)


surfactant/cmd/generate.py
```diff
@@ -223,6 +225,13 @@ def get_default_from_config(option: str, fallback: Optional[Any] = None) -> Any:
     is_eager=True,
     help="List supported input formats",
 )
+@click.option(
+    "--include_all_files",
+    is_flag=True,
+    default=False,
```
nightlark, line 231:
```suggestion
    default=get_default_from_config("include_all_files"),
```

Let's pull the default value from the `core.include_all_files` config option, so users can change the default behavior. If they haven't set anything, I think the default fallback value returned is `NoneType`, which should be considered false-y for comparisons... or `fallback=False` could be added as an additional argument to `get_default_from_config`.

surfactant/plugin/hookspecs.py
```diff
@@ -49,6 +50,7 @@ def extract_file_info(
         filetype (str): File type information based on magic bytes.
         context (Queue[ContextEntry]): Modifiable queue of entries from input config file. Existing plugins should still work without adding this parameter.
         children (List[Software]): List of additional software entries to include in the SBOM. Plugins can add additional entries, though if the plugin extracts files to a temporary directory, the context argument should be used to have Surfactant process the files instead.
+        include_all_files (bool): If all files should be included in the SBOM, rather than only those that are identified by Surfactant.
```
nightlark, line 53:
```suggestion
        include_all_files (bool): If all files should be included in the SBOM, rather than only those with types that are recognized by Surfactant. When a plugin is adding additional context entries to the queue, it should typically default to propagating this value to the new context entries that it creates.
```

I hadn't considered that info extraction plugins would care about knowing this parameter... not entirely sure about the wording here, but (in addition to propagating the value to new context entries) there are also plugins that populate a list of additional children software entries to return, and their behavior change based on this argument... I'm just not sure how they are supposed to know which ones are of types that Surfactant would have identified.

nightlark (commented):
Can we also update the documentation for config files to mention includeAllFiles? I think that is in two places:

* https://github.com/LLNL/Surfactant/?tab=readme-ov-file#build-configuration-file-for-sample
* https://github.com/LLNL/Surfactant/blob/main/docs/configuration_files.md#build-sample-configuration-file

docs/basic_usage.md
```diff
@@ -22,7 +22,8 @@ $  surfactant generate [OPTIONS] CONFIG_FILE SBOM_OUTFILE [INPUT_SBOM]
 **--recorded_institution**: (optional) the name of the institution collecting the SBOM data (default: LLNL)\
 **--output_format**: (optional) changes the output format for the SBOM (given as full module name of a surfactant plugin implementing the `write_sbom` hook)\
 **--input_format**: (optional) specifies the format of the input SBOM if one is being used (default: cytrics) (given as full module name of a surfactant plugin implementing the `read_sbom` hook)\
-**--help**: (optional) show the help message and exit
+**--help**: (optional) show the help message and exit\
+**--inclulde_all_files**: Include all files in the SBOM, rather than just those recognized by Surfactant
```
nightlark, line 26:
```suggestion
**--include_all_files**: (optional) include all files in the SBOM, rather than just those recognized by Surfactant
```

Just a stray character, and altering format some to match other entries (maybe including `(optional)` for all the flags is redundant).

nightlark (approved):

Issue #250, opened 2024-08-27
Surfactant will stall forever on mounted NTFS volumes which have named pipes


mcutshaw:
**Describe the bug**
Currently if a named pipe exists on a mounted NTFS volume Surfactant will try to read it, which will result in it blocking forever.

The issue would appear to be that we do not call `os.path.isfile` before `if ftype := pm.hook.identify_file_type(filepath=filepath):` unless the file is also a symbolic link.

**To Reproduce**
1. Mount a NTFS volume which contains a named pipe on Linux.
2. Run surfactant generate on that path.
3. Wait... Forever.

**Expected behavior**
We should ensure that we're either dealing with a regular file or a symlink to a regular file before trying to read it.

**System Information (please complete the following information):**
 - OS: Ubuntu 22.04

PR #251, opened 2024-08-28
249 250 other staging

### LINKED ISSUE #249: Unhandled exception if asm_info.PublicKey is None
**Describe the bug**
Within the pe_file.py if certain fields returned by asm_info are None, then an exception is raised.
image

**System Information (please complete the following information):**
 - OS: Ubuntu 22.04

**Additional context**
We should potentially consider wrapping the info extractors in a try/except and handle exceptions gracefully.

mcutshaw:
### Summary

If merged this pull request will fix #249 #250 and another unhandled error in the JS info extractor.

### Proposed changes

Minor checks, and a more important one to do `if os.path.isfile(filepath):` before running any file type identification.

### DIFF
--- a/surfactant/cmd/generate.py
+++ b/surfactant/cmd/generate.py
@@ -404,26 +404,26 @@ def sbom(
                             install_path = epath + "/"
                         else:
                             install_path = None
+                        if os.path.isfile(filepath):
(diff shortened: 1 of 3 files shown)


nightlark (approved):
We should probably flag os.walk files that aren't files for someone to manually look into when creating an SBOM.

PR #261, opened 2024-09-23
Adding basic framework for the CLI, no changes to old cli. 

shaynakapadia:
### Summary
This MR adds the basic setup for the new non-interactive surfactant cli. 

- add a `surfactant cli load` command
- add a `surfactant cli save` command
- add a new subfolder surfactant/cmd/cli_commands that contains the cli classes related to each subcommand (base, save, load)
- Initial structure to provide serialization for SBOMs (right now it just saves it as json (pickling does not work right now))

### Proposed changes
The changes here will migrate the existing cli interface to the new structure. Proposed workflow is below:
```
surfactant cli load sbom.json # Loads the sbom into surfactant, surfactantn saves it in ~/.surfactant in a serialized form
surfactant cli find --containerPath=^123* # Loads from serialized form and finds subset that matches args
surfactant cli add --installPath 123/ /bin/ # Adds new install path based on containerpath
surfactant cli merge # merges changes to the subset from find back into the main sbom
surfactant cli find --uuid 123 # Find one entry to edit based on uuid
surfactant cli edit --components="IsAGRAF" # Editing an array by picking the element, this one edits a specific component in this entry
Current Value: {"name": "IsAGRAF", "Vendor": "Rockwell Collins Automation"}
New Value: {"name": "IsAGRAF", "Vendor": ["Rockwell Collins Automation"], "version": "1.2.3"} 
surfactant cli edit --name # Edit a string value
Current Value: oldname.out
New Value: 1.2.3.CPO.out
surfactant cli merge # Merge changes back into the rest of the SBOM
surfactant cli save new_sbom.json # save edited sbom to a new file
```

### DIFF
--- a/docs/cli_usage.md
+++ b/docs/cli_usage.md
@@ -91,7 +91,7 @@ Our SBOM after the `cli add` command:
 ```
 
 ## surfactant cli save
-The **cli save** command saves the edited sbom to the output file specified. 
+The **cli save** command saves the edited sbom to the output file specified.
 
 ### Example:
 ```bash
--- a/docs/cli_usage.md
+++ b/docs/cli_usage.md
@@ -5,7 +5,7 @@ Some functionality we support includes:
 - Fix up path prefixes, i.e. installPath or containerPath
 - Add relationships
 
-## surfactnat cli load
+## surfactant cli load
 The ***cli load* command loads an sbom file into the cli and serializes it for faster processing when running other cli commands. On Unix-like platforms (including macOS), the XDG directory specification is followed and the serialized sbom will be stored in `${XDG_CONFIG_HOME}/surfactant/config.toml`. If the `XDG_CONFIG_HOME` environment variable is not set, the location defaults
 to `~/.config`. On Windows, the file is stored in the Roaming AppData folder at `%APPDATA%\\surfactant\\sbom_cli`.
 
--- a/docs/cli_usage.md
+++ b/docs/cli_usage.md
@@ -5,6 +5,15 @@ Some functionality we support includes:
 - Fix up path prefixes, i.e. installPath or containerPath
 - Add relationships
 
+## surfactnat cli load
+The ***cli load* command loads an sbom file into the cli and serializes it for faster processing when running other cli commands. On Unix-like platforms (including macOS), the XDG directory specification is followed and the serialized sbom will be stored in `${XDG_CONFIG_HOME}/surfactant/config.toml`. If the `XDG_CONFIG_HOME` environment variable is not set, the location defaults
+to `~/.config`. On Windows, the file is stored in the Roaming AppData folder at `%APPDATA%\\surfactant\\sbom_cli`.
+
+### Example
+```bash
+surfactant cli load sbom.json
+```
+
 ## surfactant cli find
 The **cli find** command allows users to find specific entries within a SBOM. This will allow users to:
 - Verify entries exist within the SBOM
@@ -79,4 +88,12 @@ Our SBOM after the `cli add` command:
 "installPath": ["/bin/helpers/test.exe"],
 "containerPath": ["123/helpers/test.exe"]
 }
+```
+
+## surfactant cli save
+The **cli save** command saves the edited sbom to the output file specified. 
+
+### Example:
+```bash
+surfactant cli save new_sbom.json
 ```
--- a/surfactant/cmd/cli.py
+++ b/surfactant/cmd/cli.py
@@ -5,8 +5,7 @@ from pathlib import Path
 import click
 from loguru import logger
 
-# from surfactant.cmd.cli_commands.cli_load import Load
-from surfactant.cmd.cli_commands import *
+from surfactant.cmd.cli_commands import Load, Save
 from surfactant.plugin.manager import find_io_plugin, get_plugin_manager
 from surfactant.sbomtypes._relationship import Relationship
 from surfactant.sbomtypes._sbom import SBOM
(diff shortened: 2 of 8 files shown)


surfactant/cmd/cli.py
```diff
@@ -5,12 +5,26 @@
...
 from surfactant.sbomtypes._relationship import Relationship
 from surfactant.sbomtypes._sbom import SBOM
 from surfactant.sbomtypes._software import Software
 
 
+@click.argument("sbom", type=click.File("r"), required=True)
+@click.option(
+    "--input_format",
+    is_flag=False,
+    default="surfactant.input_readers.cytrics_reader",
```
nightlark, line 19: We can use the ConfigManager so that if users have set a preferred default input format it will be used instead.
```suggestion
    default=ConfigManager().get(
        "core", "input_format", fallback="surfactant.input_readers.cytrics_reader"
    ),
```

surfactant/cmd/cli.py
```diff
@@ -109,10 +123,23 @@ def add(sbom, output, output_format, input_format, **kwargs):
...
-def edit(sbom, output_format, input_format, **kwargs):
+def handle_cli_edit(sbom, output_format, input_format, **kwargs):
     "CLI command to edit specific entry(s) in a supplied SBOM"
 
 
+@click.argument("outfile", type=click.File("w"), required=True)
+@click.option(
+    "--output_format",
+    is_flag=False,
+    default="surfactant.output.cytrics_writer",
```
nightlark, line 134: Use the ConfigManager here as well to check if the user has set a preferred default output format.

```suggestion
    default=ConfigManager().get(
        "core", "output_format", fallback="surfactant.output.cytrics_writer"
    ),
```

nightlark (commented):
It will be interesting to see how the performance for (de)serializing larger SBOMs is -- it looks like that will need to happen for every command that gets ran?

shaynakapadia:
> It will be interesting to see how the performance for (de)serializing larger SBOMs is -- it looks like that will need to happen for every command that gets ran?

Ran some timing on the `surfactant cli load` cmd, which both serializes and deserializes. Not sure why 72.7 MB and the 134 MB is going slower than the larger ones, but could be the nesting or something. Right now the serialization isn't really serialization, its just writing json to file. I was running into issues with python pickle, so am working on figuring out a workaround. 

|SBOM Size | Avg Time|
|-----|-----|
| 134 KB | 0.720 sec|
| 783 KB| 0.708 sec |
|5.8 MB|1.356 sec|
|11.5 MB| 3.641 sec|
|72.7 MB|2.806 sec|

docs/cli_usage.md
```diff
@@ -5,6 +5,15 @@ Some functionality we support includes:
 - Fix up path prefixes, i.e. installPath or containerPath
 - Add relationships
 
+## surfactnat cli load
```
nightlark, line 8:
```suggestion
## surfactant cli load
```

surfactant/cmd/cli_commands/cli_base.py
```diff
@@ -0,0 +1,88 @@
...
+    sbom_filename: str
+    subset_filename: str
+    match_functions: dict
+    camel_case_conversions: dict
+
+    def __init__(self):
+        self.sbom_filename = "sbom_cli"
+        self.subset_filename = "subset_cli"
+        # Create data directory
+        self.data_dir = self._get_cli_sbom_dir()
```
nightlark, line 38: Can the function in the ConfigManager (https://github.com/LLNL/Surfactant/blob/main/surfactant/configmanager.py#L145) get used instead? I think the one difference is it uses Local AppData on Windows -- it doesn't get synchronized between multiple systems, which is probably okay (esp. if the SBOM happens to be very big).

nightlark (commented):
nit: cli_save.py and cli_load.py are using a different docstring style than the other files.

surfactant/cmd/cli_commands/cli_base.py
```diff
@@ -0,0 +1,88 @@
...
+    """
+    A base class that implements the surfactant cli basic functionality
+
+    Attributes:
+    sbom                    An internal record of sbom entries the class adds to as it finds more matches.
+    subset                  An internal record of the subset of sbom entries from the last cli find call.
+    sbom_filename:          A string value of the filename where the loaded sbom is stored.
+    subset_filename:        A string value of the filename where the current subset result from the "cli find" command is stored.
+    match_functions         A dictionary of functions that provide matching functionality for given SBOM fields (i.e. uuid, sha256, installpath, etc)
+    camel_case_conversions  A dictionary of string conversions from all lowercase to camelcase. Used to convert python click options to match the SBOM attribute's case
```
nightlark, line 23:
```suggestion
        sbom: An internal record of sbom entries the class adds to as it finds more matches.
        subset: An internal record of the subset of sbom entries from the last cli find call.
        sbom_filename: A string value of the filename where the loaded sbom is stored.
        subset_filename: A string value of the filename where the current subset result from the "cli find" command is stored.
        match_functions: A dictionary of functions that provide matching functionality for given SBOM fields (i.e. uuid, sha256, installpath, etc)
        camel_case_conversions: A dictionary of string conversions from all lowercase to camelcase. Used to convert python click options to match the SBOM attribute's case
```

surfactant/cmd/cli_commands/cli_base.py
shaynakapadia: Got it updated! If it needs to be changed at any point, we can integrate a new location into the ConfigManger then

surfactant/cmd/cli_commands/cli_base.py
```diff
@@ -0,0 +1,79 @@
...
+        # Create data directory
+        self.data_dir = ConfigManager().get_data_dir_path()
+        self.data_dir.mkdir(parents=True, exist_ok=True)
+
+    @staticmethod
+    def serialize(sbom: SBOM) -> str:
+        """Serializes a given sbom.
+
+        Args:
+            bom (SBOM): Aninstance of an SBOM to serialize
```
shaynakapadia, line 44:
```suggestion
            bom (SBOM): An instance of an SBOM to serialize
```

nightlark (approved):

PR #262, opened 2024-09-30
Add include/exclude extensions options

### LINKED ISSUE #239: Add file extension identification config options
Add config options for the file type id based on extension to:

- Recognize a list of additional extensions
- Ignore extensions that would typically be recognized

KendallHarterAtWork:
### Summary

If merged this pull request will add options for having lists of file extensions to include and exclude.  This resolves #239.

### Proposed changes

Adds includeFileExts and excludeFileExts options to the configuration file which are lists of file extensions to forcibly include/exclude.

### DIFF
--- a/surfactant/cmd/generate.py
+++ b/surfactant/cmd/generate.py
@@ -484,10 +484,18 @@ def sbom(
                             continue
 
                         if os.path.isfile(filepath):
+                            if not entry.includeFileExts:
+                                entry.includeFileExts = []
+                            if not entry.excludeFileExts:
+                                entry.excludeFileExts = []
                             if (
                                 ftype := pm.hook.identify_file_type(filepath=filepath)
                                 or include_all_files
-                            ):
+                                or os.path.splitext(filepath)[1].lower()
+                                in [ext.lower() for ext in entry.includeFileExts]
+                            ) and os.path.splitext(filepath)[1].lower() not in [
+                                ext.lower() for ext in entry.excludeFileExts
+                            ]:
                                 try:
                                     sw_parent, sw_children = get_software_entry(
                                         context,
--- a/README.md
+++ b/README.md
@@ -142,10 +142,12 @@ In order to test out surfactant, you will need a sample file/folder. If you don'
 
 A configuration file for a sample contains the information about the sample to gather information from. Example JSON sample configuration files can be found in the examples folder of this repository.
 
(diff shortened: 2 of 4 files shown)


surfactant/cmd/generate.py
```diff
@@ -431,10 +431,16 @@ def sbom(
                                 except Exception as e:
                                     raise RuntimeError(f"Unable to process: {filepath}") from e
 
+                        if not entry.includeFileExts:
+                            entry.includeFileExts = []
+                        if not entry.excludeFileExts:
+                            entry.excludeFileExts = []
```
nightlark, line 437: I was debating if the check should happen in `surfactant/filetypeid/id_extension.py` -- but I think this is the only place that makes sense, since it needs to be able to exclude files by extension even if they were identified by something like magic bytes.

We could perhaps expand `Exts` to `Extensions`, though `Ext` is a common enough abbreviation.

surfactant/cmd/generate.py
```diff
@@ -431,10 +431,16 @@ def sbom(
...
+
                         if (
                             ftype := pm.hook.identify_file_type(filepath=filepath)
                             or include_all_files
-                        ):
+                            or os.path.splitext(filepath)[1] in entry.includeFileExts
+                        ) and os.path.splitext(filepath)[1] not in entry.excludeFileExts:
```
nightlark, line 443:
```suggestion
                            or filepath.suffix in entry.includeFileExts
                        ) and filepath.suffix not in entry.excludeFileExts:
```

`filepath` should be of type `pathlib.Path`, so we can use `.suffix` to get the extension.

nightlark: Nevermind, looks like `filepath` is set from a call to `as_posix()` so its actually a `str`.

surfactant/cmd/generate.py
nightlark: The only other thought I have about this is should we add a core config setting that lets a user set a default list of include/exclude file exts?

nightlark (approved):

surfactant/cmd/generate.py
KendallHarterAtWork: Yeah, since a file could be identified using magic bytes I'm not sure where else it could go.
I think adding a core setting in addition to the list could work, but then you run into if it overrides or adds to the lists, which may cause confusion.

PR #267, opened 2024-10-14
Native library detection plugin

wangmot:
### Summary

If merged this pull request will detect native libraries from files. It will find matches either through file name or file content using regex patterns. Statically linked libraries will also be detected. 

First, run `get_emba_db.py` to generate the EMBA database. 

Then, just run surfactant and in the output file there will be a `nativeLibraries: []` for each file that states all the libraries that it was able to detect. The output will have either `isLibrary: []` meaning the file is the library, or `containsLibrary: []` which means the libraries were statically linked within the file.


scripts/native_libraries/get_emba_db.py
```diff
@@ -0,0 +1,80 @@
+import json
+
+import requests
+
```
nightlark, line 4:
```suggestion

from surfactant.configmanager import ConfigManager
```

scripts/native_libraries/get_emba_db.py
```diff
@@ -0,0 +1,80 @@
...
+                        "filecontent": [filecontent],
+                    }
+                else:
+                    database[lib_name]["filecontent"].append(filecontent)
+
+    return database
+
+
+url = "https://raw.githubusercontent.com/e-m-b-a/emba/master/config/bin_version_strings.cfg"
+json_file_path = "/Users/tenzing1/surfactant_new_venv/Surfactant/surfactant/infoextractors/native_lib_patterns.json"
```
nightlark, line 59:
```suggestion
json_file_path = ConfigManager().get_data_dir_path() / "native_lib_patterns"/ "emba.json"
```

Getting rid of the hard-coded path (+ your username) -- use a path that conforms to XDG base directory specification and handles differences in location for different operating systems.

scripts/native_libraries/get_emba_db.py
```diff
@@ -0,0 +1,80 @@
+import json
```
nightlark, line 1:
```suggestion
import json
import os
```

scripts/native_libraries/get_emba_db.py
```diff
@@ -0,0 +1,80 @@
...
+        if filecontent_list[i].startswith("^"):
+            filecontent_list[i] = filecontent_list[i][1:]
+
+        if filecontent_list[i].endswith("\\$"):
+            pass
+        else:
+            if filecontent_list[i].endswith("$"):
+                filecontent_list[i] = filecontent_list[i][:-1]
+
+with open(json_file_path, "w") as json_file:
```
nightlark, line 79:
```suggestion
os.makedirs(os.path.dirname(json_file_path), exist_ok=True)
with open(json_file_path, "w") as json_file:
```

Making sure the directories needed to contain the file pointed to by json_file_path exist.

surfactant/infoextractors/native_lib_file.py
```diff
@@ -0,0 +1,79 @@
+import json
+import pathlib
+import re
+from typing import Any, Dict, List
+
+from loguru import logger
+
+import surfactant.plugin
+from surfactant.sbomtypes import SBOM, Software
```
nightlark, line 9:
```suggestion
from surfactant.sbomtypes import SBOM, Software
from surfactant.configmanager import ConfigManager
```

surfactant/infoextractors/native_lib_file.py
```diff
@@ -0,0 +1,79 @@
...
+
+    found_libraries = set()
+
+    # Match based on filename
+    filenames_list = match_by_attribute("filename", filename, database)
+    if len(filenames_list) > 0:
+        for match in filenames_list:
+            library_name = match["library"]
+            if library_name not in found_libraries:
+                native_lib_info["nativeLibraries"].append(match)
```
nightlark, line 43: I'm wondering if this should use a different key like "isNativeLibrary" -- theoretically a file name matching would imply that the file itself is that library/binary.

e.g. A file could have a copy of OpenSSL embedded in it (contains the native library based on a pattern match), but if the file name matches "libopenssl" it would be better described as "being" OpenSSL (which is a more accurate description than just saying it "contains" OpenSSL). It is also unlikely that a file will match patterns for multiple different things (if it does, then those file name patterns are probably not specific enough).

nightlark (commented):
Looks pretty good, mostly a few changes related to where the database of patterns is stored.

Let's add this helper function to the ConfigManager class (in `surfactant/configmanager.py`):
```    def get_data_dir_path(self) -> Path:
        """Determines the path to the data directory, for storing things such as databases.

        Returns:
            Path: The path to the data directory.
        """
        if platform.system() == "Windows":
            data_dir = Path(os.getenv("LOCALAPPDATA", os.path.expanduser("~\\AppData\\Local")))
        else:
            data_dir = Path(os.getenv("XDG_DATA_HOME", os.path.expanduser("~/.local/share")))
        data_dir = data_dir / self.app_name
        return data_dir
```

surfactant/infoextractors/native_lib_file.py
```diff
@@ -0,0 +1,89 @@
...
+import re
+from typing import Any, Dict, List
+
+from loguru import logger
+
+import surfactant.plugin
+from surfactant.configmanager import ConfigManager
+from surfactant.sbomtypes import SBOM, Software
+
+
```
nightlark, line 11:
```suggestion


def load_pattern_db():
    # Load regex patterns into database var
    try:
        with open(native_lib_patterns, "r") as regex:
            database = json.load(regex)
            return database
    except FileNotFoundError:
        logger.warning(f"File not found for native library detection: {native_lib_patterns}")
        return None
        
        
# Load the pattern database once at module import
database = load_pattern_db()


```

surfactant/infoextractors/native_lib_file.py
```diff
@@ -0,0 +1,89 @@
...
+    native_lib_info: Dict[str, Any] = {"nativeLibraries": []}
+    native_lib_patterns = ConfigManager().get_data_dir_path() / "native_lib_patterns" / "emba.json"
+
+    # Load regex patterns into database var
+    try:
+        with open(native_lib_patterns, "r") as regex:
+            database = json.load(regex)
+    except FileNotFoundError:
+        logger.warning(f"File not found: {native_lib_patterns}")
+        return None
```
nightlark, line 33:
```suggestion
    # Check if the pattern database loaded successfully
    if not database:
        return None
```

nightlark:
Interesting, I came across two patterns that don't compile and cause an error:

- `C++\ RTMP\ Server\ .*\ version\ v[0-9](\.[0-9])+?\ r\.[0-9]+` due to the `++` at the start being intended to match literally; emba may be taking advantage of some non-standard behavior in the tool they use for regex parsing, whereas Python's `re` module (and every other regexp parser on https://regex101.com/) either doesn't like that construct and complains about the pattern, or treats it as matching multiple C's.
- `User-Agent: Siemens Canada Limited - ROX2 - [0-9](\.[0-9]+)?+$` is disliked by Python's `re` module due to the `?+` towards the end. Other regexp parsers give an error, or treat it as basically the same as just having a `?`... I'm not sure if it intends to match against a literal `+` or not.. or maybe they meant to just have `+` at the end so it matches e.g. `4.2.4.4` with an arbitrary number of dots, instead of just a maximum of one dot in the version number.

Will need to test whatever they are using for matching with the regular expressions, but it is possible the first doesn't have the behavior they intended (and the 2nd is just ambiguous).

scripts/native_libraries/get_emba_db.py
```diff
@@ -0,0 +1,86 @@
...
+                        "filecontent": [],
+                    }
+            else:
+                if lib_name not in database:
+                    database[lib_name] = {
+                        "filename": name_patterns,
+                        "filecontent": [filecontent],
+                    }
+                else:
+                    database[lib_name]["filecontent"].append(filecontent)
```
nightlark, line 59:
```suggestion
                try:
                    if lib_name not in database:
                        database[lib_name] = {
                            "filename": name_patterns,
                            "filecontent": [filecontent],
                        }
                    else:
                        database[lib_name]["filecontent"].append(filecontent)
                except re.error as e:
                    print(f"Error parsing file content regexp {filecontent}: {e}")
```

When constructing the database, let's catch the patterns that cause issues to make sure when we go to generate the SBOM we don't run into invalid regexp patterns.

nightlark: We could consider doing something similar for file name patterns.

nightlark:
I think they are using `grep -o -a -E "<pattern>"` for matching, and for both of those patterns grep says `repetition-operator operand invalid`. So I think these are actually just broken patterns in the EMBA database.

scripts/native_libraries/get_emba_db.py
wangmot: Hmm I think im crazy cause both of the above patterns compile for me. I tested it using `re.compile` as well. I'll rerun again tomorrow!

Also, regarding doing the `re` check for the file name patterns, I think in this case it won't be necessary? Cause EMBA does not have file name patterns, and if we see that `fields[1] == "strict"` then we just set the `lib_name` under the file name pattern section. So, we wouldn't have any real patterns in the file name patterns since the pattern would just be a string literal of the file name.

nightlark: Try using an older version of Python (on LC cluster such as dane.llnl.gov, `module spider python` to list available Python versions then `module spider python/3.9.12` to "load" the different version). What I'm finding is that Python <=3.10 gives the multiple repeat error, 3.11 outputs nothing, and 3.12 prints a SyntaxWarning about `\.` being an invalid escape sequence.

Since users are going to be getting the database file themselves, presumably with the same version of Python version that they will run Surfactant with, the check should catch patterns that break on their Python version. 

---

In match by attribute we're still using re.search to match the file name using a regular expression and characters such as + and ? are valid in file names, so there could be a file name that fails to compile with the `re` module:

```python
if attribute == "filename":
    matches = re.search(pattern, content)
```

If the file names in the database will always be exact matches rather than patterns, we can just do a regular string comparisons instead of pattern matching.

wangmot: Okay for filename I switched it to do a regular string comparison. Content is basically filename below:
`if name == content:`
    `libs.append({"isLibrary": name})`
    
EMBA db has about 806 libraries total, and only ~37 have the "strict" keyword. These are the ones where we set the `lib_name` to filename since EMBA does not have file content patterns for the `strict` keyword. And the filenames don't have any special characters so simple string comparison should work!

surfactant/infoextractors/native_lib_file.py
```diff
@@ -96,7 +96,7 @@ def match_by_attribute(attribute: str, content: str, patterns_database: Dict) ->
                 if attribute == "filename":
                     if name == content:
```
nightlark, line 97:
```suggestion
                    if name.lower() == content.lower():
```

Let's make this a case-insensitive comparison.

surfactant/infoextractors/native_lib_file.py
```diff
@@ -0,0 +1,104 @@
...
+            for pattern in library[attribute]:
+                if attribute == "filename":
+                    if name.lower() == content.lower():
+                        libs.append({"isLibrary": name})
+
+                elif attribute == "filecontent":
+                    matches = re.search(pattern.encode("utf-8"), content)
+                    if matches:
+                        libs.append({"containsLibrary": name})
+    return libs
```
nightlark, line 104:
```suggestion
    for lib_name, lib_info in patterns_database.items():
        if attribute in library:
            for pattern in lib_info[attribute]:
                if attribute == "filename":
                    if pattern.lower() == content.lower():
                        libs.append({"isLibrary": lib_name})

                elif attribute == "filecontent":
                    matches = re.search(pattern.encode("utf-8"), content)
                    if matches:
                        libs.append({"containsLibrary": lib_name})
    return libs
```

Renaming a few variables to make it a bit clearer what they are. Also using "pattern" (the filecontent list) for the file name check -- while it is currently just a single entry that matches the lib_name, theoretically other pattern databases could have multiple file names lited.

nightlark (approved):

PR #279, opened 2024-11-04
Added plugin list subcommand functionality

willis89pr:
### Summary

### DIFF
--- a/surfactant/cmd/plugin.py
+++ b/surfactant/cmd/plugin.py
@@ -1,0 +1,13 @@
+import click
+from surfactant.plugin.manager import get_plugin_manager
+
+@click.command("list")
+def list():
+    pm = get_plugin_manager()
+    print("-------")
+    print("PLUGINS")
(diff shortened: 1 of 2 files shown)


surfactant/cmd/plugin.py
```diff
@@ -0,0 +1,15 @@
+import click
+
+from surfactant.plugin.manager import get_plugin_manager
+
+
+@click.command(name="list")
+def display():
```
nightlark, line 7: Let's also update the function name from `display` to something more specific like `plugin_list_cmd`.

Also let's add a docstring as a line below it with the `"""` syntax (similar to the "Manage plugins" one added for the plugin command group) so that `surfactant plugin --help` will show a brief description of what the list subcommand does.

willis89pr: Thank you. I made the changes.

nightlark (approved):

PR #281, opened 2024-11-11
CYT-1123 adds enable and disable to plugins command

willis89pr:
### Summary

If merged this pull request adds plugin enable and plugin disable subcommands to enable and disable plugins. It also updates the list sub-command to list disabled plugins.

### DIFF
--- a/surfactant/cmd/plugin.py
+++ b/surfactant/cmd/plugin.py
@@ -1,9 +1,7 @@
 import click
-
 from surfactant.plugin.manager import get_plugin_manager
 
-
-@click.command("display")
+@click.command(name="list")
 def display():
     pm = get_plugin_manager()
     print("-------")
--- a/surfactant/cmd/plugin.py
+++ b/surfactant/cmd/plugin.py
@@ -2,7 +2,8 @@ import click
 from surfactant.plugin.manager import get_plugin_manager
 
 @click.command(name="list")
-def display():
+def plugin_list_cmd():
+    """Lists plugins."""
     pm = get_plugin_manager()
     print("-------")
     print("PLUGINS")
--- a/surfactant/cmd/plugin.py
+++ b/surfactant/cmd/plugin.py
@@ -2,6 +2,7 @@ import click
 
 from surfactant.plugin.manager import get_plugin_manager
 
+
 @click.command("display")
 def display():
     pm = get_plugin_manager()
--- a/surfactant/cmd/plugin.py
+++ b/surfactant/cmd/plugin.py
@@ -23,16 +23,18 @@ def plugin_disable_cmd(plugin_names):
     """Disables one or more plugins."""
     if not plugin_names:
         raise click.UsageError("At least one plugin name must be specified.")
-
+    section = 'core'
+    section_key = 'blocked'
+    
     pm = get_plugin_manager()
     config_manager = ConfigManager()
 
     # Debugging: Print config file path and current config
-    print(config_manager._get_config_file_path())
-    config_manager.print_config()
+    #print(config_manager._get_config_file_path())
+    #config_manager.print_config()
 
     # Retrieve the current list of blocked plugins
-    current_blocked_plugins = config_manager.get('plugins', 'blocked', [])
+    current_blocked_plugins = config_manager.get(section, section_key, [])
 
     # Ensure current_blocked_plugins is a list
     if isinstance(current_blocked_plugins, str):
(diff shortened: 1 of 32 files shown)


surfactant/cmd/plugin.py
```diff
@@ -14,3 +15,74 @@ def plugin_list_cmd():
...
+    section_key = "disable_plugins"
+
+    # Retrieve the current list of plugins that should be blocked
+    current_blocked_plugins = config_manager.get(section, section_key, [])
+
+    print("-------")
+    print("DISABLED PLUGINS")
+    for disabled_plugin in current_blocked_plugins:
+        print("-------")
+        print(f"name: {disabled_plugin}")
```
nightlark, line 31:
```suggestion
    for disabled_plugin in current_blocked_plugins:
        print(f"\tname: {disabled_plugin}")
```
I think we can make the output a bit easier to follow with indentation

surfactant/cmd/plugin.py
```diff
@@ -14,3 +15,74 @@ def plugin_list_cmd():
         print(f"canonical name: {pm.get_canonical_name(p)}")
         plugin_name = pm.get_name(p) if pm.get_name(p) else ""
         print(f"name: {plugin_name}")
+
+    # List disabled pluginsi
```
nightlark, line 19:
```suggestion
    # List disabled plugins
```

surfactant/cmd/plugin.py
```diff
@@ -14,3 +15,74 @@ def plugin_list_cmd():
         print(f"canonical name: {pm.get_canonical_name(p)}")
         plugin_name = pm.get_name(p) if pm.get_name(p) else ""
         print(f"name: {plugin_name}")
```
nightlark, line 17:
```suggestion
        print(f"\t> canonical name: {pm.get_canonical_name(p)}")
        plugin_name = pm.get_name(p) if pm.get_name(p) else ""
        print(f"\t  name: {plugin_name}")
```
I'm wondering if maybe indenting the list of plugins and removing the `print("-------")` separators will make the output more legible.

surfactant/plugin/manager.py
```diff
@@ -65,6 +66,25 @@ def _register_plugins(pm: pluggy.PluginManager) -> None:
...
+
+        # Unregister the plugin
+        plugin = pm.unregister(name=plugin_name)
+        if plugin is None:
+            print(f"Plugin '{plugin_name}' not found.")
+            continue
+
+        # Block the plugin to prevent future registration
+        pm.set_blocked(plugin_name)
+
```
nightlark, line 87: Let's move all of this into its own disable plugins function that takes a pluggy.PluginManager as an argument, then call that function from in the get_plugin_manager function below (right after the call to _register_plugins.

surfactant/plugin/manager.py
```diff
@@ -65,6 +66,25 @@ def _register_plugins(pm: pluggy.PluginManager) -> None:
...
+    for plugin_name in current_blocked_plugins:
+        # Check if the plugin is already blocked
+        if pm.is_blocked(plugin_name):
+            print(f"Plugin '{plugin_name}' is already disabled.")
+            continue
+
+        # Unregister the plugin
+        plugin = pm.unregister(name=plugin_name)
+        if plugin is None:
+            print(f"Plugin '{plugin_name}' not found.")
```
nightlark, line 82:
```suggestion
            logger.info(f"Disabled plugin '{plugin_name}' not found.")
```

Log using logger instead of print. Actually not sure if this is a message that a user necessarily cares about either, other than for the purposes of cleaning up their disabled plugin list.

surfactant/plugin/manager.py
```diff
@@ -65,6 +66,25 @@ def _register_plugins(pm: pluggy.PluginManager) -> None:
...
         pm.register(plugin)
 
+    config_manager = ConfigManager()
+
+    # Retrieve the current list of blocked plugins
+    current_blocked_plugins = config_manager.get("core", "disable_plugins", [])
+    for plugin_name in current_blocked_plugins:
+        # Check if the plugin is already blocked
+        if pm.is_blocked(plugin_name):
+            print(f"Plugin '{plugin_name}' is already disabled.")
```
nightlark, line 76:
```suggestion
```

I don't think this is a case that users need to care about seeing.

nightlark (approved):
LGTM

The plugin listing looks a bit better -- I feel like it could still look cleaner, but I'm not sure how at the moment; if inspiration strikes later we can tweak it some in a future PR.

PR #286, opened 2024-11-18
Update README.md plugins section

willis89pr:
Adds instructions for the plugin sub-command.

### Summary

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -26,7 +26,8 @@ In order to create a plugin, you will need to write your implementation for one 
 
 ### Step 2. Write pyproject.toml File
 
-Once you have written your plugin, you will need to write a pyproject.toml file. Include any relevant project metadata/dependencies for your plugin, as well as an entry-point specification (example below) to make the plugin discoverable by surfactant. Once you write your .toml file, you can `pip install .` your plugin.
+Once you have written your plugin, you will need to write a pyproject.toml file. Include any relevant project metadata/dependencies for your plugin, as well as an entry-point specification (example below) to make the plugin discoverable by surfactant. Once you write your .toml file, you can `surfactant plugin install <plugin name>` to install your plugin. Alternatively, you can `pip install .` your plugin.
+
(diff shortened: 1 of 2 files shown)


docs/plugins.md
```diff
@@ -46,6 +47,6 @@ def write_sbom(sbom: SBOM, outfile) -> None:
 [project.entry-points."surfactant"]
 sampleplugin = "sampleplugin"
 ```
-From the same folder as your sampleplugin files, run `pip install .` to install your plugin and surfactant will automatically load and use the plugin.
+Run `surfactant plugin install sampleplugin` to install the plugin. Surfactant will automatically load and use the plugin. Alternatively, from the same folder as your sampleplugin files, run `pip install .`.
```
nightlark, line 50: Since `surfactant plugin install` is calling pip, this should probably just say to run `surfactant plugin install .` (or the path to the folder on their local filesystem that the plugin is located in) so that pip doesn't try to find a package on PyPI that either doesn't exist yet (or a completely different package someone else created that happens to have the same name).

We could mention uploading their plugin to PyPI for easier sharing and to enable installation using the PyPI package name for the plugin.

docs/plugins.md
```diff
@@ -26,7 +26,8 @@ In order to create a plugin, you will need to write your implementation for one
 
 ### Step 2. Write pyproject.toml File
 
-Once you have written your plugin, you will need to write a pyproject.toml file. Include any relevant project metadata/dependencies for your plugin, as well as an entry-point specification (example below) to make the plugin discoverable by surfactant. Once you write your .toml file, you can `pip install .` your plugin.
+Once you have written your plugin, you will need to write a pyproject.toml file. Include any relevant project metadata/dependencies for your plugin, as well as an entry-point specification (example below) to make the plugin discoverable by surfactant. Once you write your .toml file, you can `surfactant plugin install <plugin name>` to install your plugin. Alternatively, you can `pip install .` your plugin.
```
nightlark, line 29: We should say `.` or <path to plugin's folder> instead -- at this stage, they probably haven't uploaded their package to PyPI yet.

nightlark (approved):

PR #287, opened 2024-11-19
Add TUI for creating/modifying config files.

KendallHarterAtWork:
### Summary

If merged this pull request will add a command that provides a TUI for creating and editing input configuration files.

### Proposed changes

Running the following command:
```bash
surfactant config_tui input_file.json
```

Will allow the creation/editing of input_file.json


pyproject.toml
```diff
@@ -54,6 +54,7 @@ dependencies = [
     "loguru==0.7.*",
     "flask==3.*",
     "tomlkit==0.13.*",
+    "textual-dev==1.*",
```
nightlark, line 57:
```suggestion
    "textual==1.*",
```

I think this dependency should just be textual.

pyproject.toml
```diff
@@ -54,6 +54,7 @@ dependencies = [
     "loguru==0.7.*",
     "flask==3.*",
     "tomlkit==0.13.*",
+    "textual==1.*",
```
nightlark, line 57:
```suggestion
    "textual==0.88.*",
```

That's kind of unexpected -- textual version numbers don't match up with the textual-dev version numbers.

surfactant/web-files/config_tui.tcss
```diff
@@ -0,0 +1,64 @@
...
+   width: 1fr;
+}
+
+ConfigTUI #config_entries {
+   width: 9fr;
+}
+
+ConfigTUI #add_entry_scrollable {
+   width: 1fr;
+}
```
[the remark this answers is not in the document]
KendallHarterAtWork: Yeah, there are possibly some other styling rules that are out of date too so I'll go through and remove any obsolete ones.

nightlark (approved):

Issue #233, opened 2024-07-23
Recognize perl scripts


nightlark:
Add support for recognizing perl scripts -- this is most likely just adding to the list of recognized file extensions for file type id, but any other ideas for verifying that the file is actually a perl script based on contents would be nice to have.

Related file: https://github.com/LLNL/Surfactant/blob/main/surfactant/filetypeid/id_extension.py

Eriken79:
Hey @nightlark just put in a PR for this, its PR #316. Not much of a change, but hopefully it looks good! Also wrote out some detailed comments on problems with recognizing Perl scripts!

PR #316, opened 2025-01-06
Recognize Perl Scripts

Eriken79:
I added .pl and .pm file extensions to be recognized and added Perl as a recognizable interpreter

If merged this pull request will allow Surfactant to recognize files with the .pl and .pm extension and recognize Perl as an interpreter. I added .pl, .pm, to the filetype_extensions list in surfactant/filetypeid/id_extensions.py and Perl to the interpreters list.

This PR addresses issue #233 and adds the most rudimentary fix to the issue. A couple thoughts on this issue:
- I'm not sure how Perl is most commonly  distributed in the wild, but I am aware of PerlApp that is used to compile a Perl script into a standalone binary with the script embedded and a copy of the Perl interpreter. Unfortunately, my change wouldn't catch that, but my intuition is that Perl is generally just run as a script with the Perl interpreter on a .pl file which my change would catch.
- I think there are some other ways that we could identify a Perl script such as $, @, % for variable names, looking for pragmas like warnings or strict, or checking for unique built in functions like chomp, but the presence of all of those would only raise the statistical likelihood of it being a Perl file and aren't necessarily definitive.

I think this is a solid fix now and maybe if we run into problems identifying Perl files in the future we can explore this again!

### DIFF
--- a/surfactant/filetypeid/id_extension.py
+++ b/surfactant/filetypeid/id_extension.py
@@ -27,7 +27,7 @@ def identify_file_type(filepath: str) -> Optional[str]:
         ".php": "PHP",
         ".bat": "BATCH",
         ".pl": "PERL_OR_PROLOG",
-        ".pm": "PERL_MODULE"
+        ".pm": "PERL_MODULE",
     }
     _interpreters = {
         b"sh": "SHELL",
--- a/surfactant/filetypeid/id_extension.py
+++ b/surfactant/filetypeid/id_extension.py
@@ -26,6 +26,8 @@ def identify_file_type(filepath: str) -> Optional[str]:
         ".htm": "HTML",
         ".php": "PHP",
         ".bat": "BATCH",
+        ".pl": "PERL",
+        ".pm": "PERL_MODULE"
     }
     _interpreters = {
         b"sh": "SHELL",
@@ -34,6 +36,7 @@ def identify_file_type(filepath: str) -> Optional[str]:
         b"php": "PHP",
         b"python": "PYTHON",
         b"python3": "PYTHON",
+        b"perl": "PERL",
     }
     try:
         with open(filepath, "rb") as f:
--- a/surfactant/filetypeid/id_extension.py
+++ b/surfactant/filetypeid/id_extension.py
@@ -26,7 +26,7 @@ def identify_file_type(filepath: str) -> Optional[str]:
         ".htm": "HTML",
         ".php": "PHP",
         ".bat": "BATCH",
-        ".pl": "PERL",
+        ".pl": "PERL_OR_PROLOG",
         ".pm": "PERL_MODULE"
     }
     _interpreters = {


surfactant/filetypeid/id_extension.py
```diff
@@ -26,6 +26,8 @@ def identify_file_type(filepath: str) -> Optional[str]:
         ".htm": "HTML",
         ".php": "PHP",
         ".bat": "BATCH",
+        ".pl": "PERL",
+        ".pm": "PERL_MODULE",
```
nightlark, line 30: If we start trying to parse perl scripts to try to extract some info on what they're doing, is there any difference in how we'd want to treat .pl files vs .pm files?

Eriken79: Pm files don't start with a #! which is probably an important identifier. They start with `package <name>;` and then in the .pl file that uses the .pm, there is a top-level `use <pm-name>; `line to import it. Then the .pl file that uses the .pm will call subroutines (functions) from it like a namespace in C++. .pm files can also use other .pm files so there is a dependency hierarchy there. Other than the fact that .pm's are always dependent on .pl's the syntax is identical. From what I understand, you would need to understand the dependency of .pm's on a .pl file if you wanted to recover the semantic information for the subroutines.

nightlark:
If there's anything in those file types that we can mostly guarantee will be present that can distinguish them from a perl script (ideally at a fixed location, and near the start of a file).

nightlark:
Maybe we'd want to rely on the #! line (if present) to distinguish between perl and prolog. Linguist apparently also has trouble distinguishing between whether a .pl file is prolog or perl when a #! is not present, and from my understanding of some of the comments using the presence of certain keywords didn't really help: https://github.com/github-linguist/linguist/issues/435

Another thought it perhaps we could recognize there will be ambiguity and return PERL_OR_PROLOG.

Eriken79:
After doing some research, it doesn't seem like there's anything in IRAF or Prolog that is easily identifiable in the header. That being said, I definitely think the #! with Perl as the interpreter is a pretty definitive identifier, so returning ambiguity with the filetype_extensions list might be a good idea. I think Perl and Prolog files are fairly easily identifiable from each other if the body of the code is examined. For continuity, do we want to have ambiguity show up in the id_extension.py file and then later downstream we clear that ambiguity by examining the body? Also, do we want to have IRAF in that, returning PERL_OR_PROLOG_OR_IRAF?

nightlark:
We could either try to clear it up with a check for the #! in id_extension.py, or just say since there's the potential ambiguity anyway if there is no #! (or its one we don't recognize), it is up to any info extraction plugin to make its own call. I guess any info extractor should in theory already be written in a way that is tolerant to malformed files, so for a perl info extraction plugin, a file with prolog code would be a malformed perl file...

I'd be inclined to leave out IRAF, since that seems a bit of an obscure format, and I don't imagine it has any information we'd want to extract for SBOMs.

Eriken79:
Hmm, I'm assuming adding `perl` in the _interpreters list checks for the #!/bin/perl? If so, that's present in my merge request. Are you saying then we shouldn't catch ".pl" in _filetype_extensions Or it should just return an ambiguous value like PERL_OR_PROLOG? Or possibly not catch anything in id_extensions.py and leave it up to the info extractor plugins/#! extraction?

nightlark:
You're right, based on the order of checks it looks for the #! first, so if it specifies perl as the interpreter then it will return PERL. Given that the file extension is the 2nd check, let's make it return PERL_OR_PROLOG for .pl files.

Eriken79:
Alright, updated .pl extension to be PERL_OR_PROLOG! Let me know if there's anything else to consider for this PR!

nightlark (approved):

PR #317, opened 2025-01-07
Add check for section header string table in ELF infoextractor

JosephHAtWork:
### Summary

If merged this pull request adds a check to the ELF file info-extractor that will skip the collection of section information (and avoid raising a `ELFParseError: String Table not found` in the process) on binaries that have a malformed or missing section header string table.

### DIFF
--- a/surfactant/infoextractors/elf_file.py
+++ b/surfactant/infoextractors/elf_file.py
@@ -80,81 +80,83 @@ def extract_elf_info(filename: str) -> object:
         # Get a human readable name for the OS
         file_details["OS"] = _EI_OSABI_NAME.get(file_details["elfOsAbi"], "")
 
-        for section in elf.iter_sections():
-            if section.name == ".interp":
-                file_details["elfInterpreter"].append(section.data().rstrip(b"\x00").decode())
-            if section.name == ".comment":
-                for v in section.data().rstrip(b"\x00").split(b"\x00"):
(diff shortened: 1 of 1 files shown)


nightlark:
The pre-commit check has some errors that should be fixed.

surfactant/infoextractors/elf_file.py
```diff
@@ -79,82 +80,84 @@ def extract_elf_info(filename: str) -> object:
...
+                            dt_flags_1_entry["DF_1_ORIGIN"] = bool(
+                                tag.entry.d_val & ENUM_DT_FLAGS_1["DF_1_ORIGIN"]
+                            )
+                            # Ignore the default library search path
+                            dt_flags_1_entry["DF_1_NODEFLIB"] = bool(
+                                tag.entry.d_val & ENUM_DT_FLAGS_1["DF_1_NODEFLIB"]
+                            )
+                            file_details["elfDynamicFlags1"].append(dt_flags_1_entry)
+        except ELFParseError as e:
+            logger.warning(f"Error while parsing ELF sections: {repr(e)}")
```
nightlark, line 160:
```suggestion
            logger.warning(f"Error while parsing ELF sections in {filename}: {repr(e)}")
```

We should also make sure it is clear which file caused the parsing errors.

nightlark (approved):

PR #318, opened 2025-01-07
Add initialization hook to plugin system

willis89pr:
### Summary

This change improves resource management and prepares the system for future enhancements that require plugin-specific setup.

### Proposed changes

- Introduced a new `init_hook` in `hookspecs.py` to allow plugins to perform initialization tasks, such as loading databases or setting up resources.
- Implemented the `init_hook` in the `js_file` plugin to load the JavaScript library database during initialization.
- Updated the plugin manager to call the `init_hook` using a helper function, ensuring plugins are initialized based on the command context.
- Enhanced modularity and efficiency of the plugin system by allowing conditional initialization.


surfactant/plugin/manager.py
```diff
@@ -192,3 +192,20 @@ def find_plugin_by_name(pm: pluggy.PluginManager, name: str):
...
+        hook_filter: A list of hook names to filter which plugins get initialized.
+        command_name: The name of the command invoking the initialization.
+    """
+    for plugin in pm.get_plugins():
+        if is_hook_implemented(pm, plugin, "init_hook"):
+            # Check if the plugin implements any of the hooks in the filter
+            if hook_filter:
+                if not any(is_hook_implemented(pm, plugin, hook) for hook in hook_filter):
+                    continue
+            plugin.init_hook(command_name=command_name)
```
nightlark, line 211: Probably okay to fallback to initializing everything if no hook_filter is provided -- it will just likely be more things happening than needed if we ever leave it unset (probably only a very limited number of external users would be interested in this function, maybe pEyeOn or similar that tries to use some of our plugins).

surfactant/infoextractors/js_file.py
```diff
@@ -126,20 +152,10 @@ def short_name():
...
-
-
-js_lib_database = load_db()
+@surfactant.plugin.hookimpl
+def init_hook(command_name=None):
+    """Initialization hook to load the JavaScript library database."""
+    if command_name != "update-db":  # Do not load the database if only updating the database.
+        click.echo("Initializing js_file...")
+        js_db_manager.load_db()
+        click.echo("Initializing js_file complete.")
```
nightlark, line 161: The echoes could be logger messages instead -- maybe set to INFO, or DEBUG (to reduce output except when debugging issues)?

nightlark (approved):

surfactant/plugin/manager.py
willis89pr: Wouldn't that make the use of hook_filter non-functional?

surfactant/plugin/manager.py
```diff
@@ -192,3 +192,20 @@ def find_plugin_by_name(pm: pluggy.PluginManager, name: str):
             return plugin
 
     return None
+
+
+def call_init_hooks(pm, hook_filter=None, command_name=None):
```
nightlark, line 197:
```suggestion
def call_init_hooks(pm: pluggy.PluginManager, hook_filter: List[str]=None, command_name: str=None):
```

surfactant/infoextractors/js_file.py
```diff
@@ -126,20 +151,10 @@ def short_name():
...
-        logger.warning(
-            "Javascript library pattern database database could not be loaded. Run `surfactant plugin update-db js_file` to fetch the pattern database."
-        )
-        return None
-    return database
-
-
-js_lib_database = load_db()
+@surfactant.plugin.hookimpl
+def init_hook(command_name=None):
```
nightlark, line 155:
```suggestion
def init_hook(command_name: Optional[str]=None):
```

surfactant/infoextractors/js_file.py
```diff
@@ -1,12 +1,11 @@
 # Copyright 2023 Lawrence Livermore National Security, LLC
-# SEe the top-level LICENSE file for details.
+# See the top-level LICENSE file for details.
 #
 # SPDX-License-Identifier: MIT
 import json
 import re
 from typing import Any, Dict, List
```
nightlark, line 7:
```suggestion
from typing import Any, Dict, List, Optional
```

surfactant/infoextractors/js_file.py
```diff
@@ -15,6 +14,31 @@
...
+        try:
+            with open(js_lib_file, "r") as regex:
+                self.js_lib_database = json.load(regex)
+        except FileNotFoundError:
+            logger.warning(
+                "Javascript library pattern database could not be loaded. Run `surfactant plugin update-db js_file` to fetch the pattern database."
+            )
+            self.js_lib_database = None
+
+    def get_database(self):
```
nightlark, line 35:
```suggestion
    def get_database(self) -> Optional[object]:
```

Not sure what the right type annotation for this would be... probably just an object.

willis89pr: for Dict[str, Any], the str is the key and Any is anything.

nightlark: Yes, if the JSON file we load is in the format:
```json
{
  ...
}
```

But other things like this are also valid JSON files that wouldn't be a Dict[str, Any]:
```json
[
  ...
]
```

surfactant/plugin/manager.py
willis89pr: Okay, I think it is the way its supposed to. if None/[], then init. If any of the filters are implemented then, init. If none of the filters are implemented, then skip.

nightlark (approved):

PR #320, opened 2025-01-13
Refactor and Relocate Native Library Pattern Loading

willis89pr:
### Summary

This pull request refactors the native library pattern loading logic by introducing a new class and relocates the script to a more appropriate module.

### Proposed changes

Refactoring:
        Introduced the NativeLibDatabaseManager class to encapsulate the database loading logic, enhancing code organization and maintainability.
        Moved the database loading logic from a module-level function to a class method for better encapsulation.

Relocation:
        Moved the native_libraries.get_emba_db.py script to the infoextractors.native_lib_file.py module, consolidating related functionality.

Initialization Hook:
        Added an init_hook to initialize the database unless the command is "update-db", providing better control over the initialization process.
        Improved logging to offer clear feedback during the initialization phase.


surfactant/infoextractors/native_lib_file.py
```diff
@@ -102,3 +109,121 @@ def match_by_attribute(attribute: str, content: str, patterns_database: Dict) ->
...
+
+    if response.status_code == 404:
+        logger.error("Resource not found.")
+    else:
+        logger.error("An error occurred.")
+
+    return None
+
+
+def parse_cfg_file(content):
```
nightlark, line 129:
```suggestion
def parse_emba_cfg_file(content: str) -> Dict[str, Dict[str, List[str]]]:
```

Since we could potentially add other sources for the database beyond emba, let's rename the function. And I think the return type is something like this...

surfactant/infoextractors/native_lib_file.py
```diff
@@ -102,3 +109,121 @@ def match_by_attribute(attribute: str, content: str, patterns_database: Dict) ->
...
+                    logger.error(f"Error parsing file content regexp {filecontent}: {e}")
+
+    return database
+
+
+@surfactant.plugin.hookimpl
+def update_db() -> str:
+    file_content = download_database()
+    if file_content is not None:
+        parsed_data = parse_cfg_file(file_content)
```
nightlark, line 187:
```suggestion
        parsed_data = parse_emba_cfg_file(file_content)
```

surfactant/infoextractors/native_lib_file.py
```diff
@@ -102,3 +109,121 @@ def match_by_attribute(attribute: str, content: str, patterns_database: Dict) ->
...
+
+        # Create a dictionary for this entry and add it to the database
+        # Strict mode is deprecated so those entries will be matched just by filename
+        if fields[1] == "" or fields[1] == "strict":
+            if fields[1] == "strict":
+                if lib_name not in database:
+                    database[lib_name] = {
+                        "filename": [lib_name],
+                        "filecontent": [],
+                    }
```
nightlark, line 166: Unlikely to happen in this dataset, but I guess this should have an else case that does `database[lib_name]["filename"].append(lib_name)`

surfactant/infoextractors/native_lib_file.py
```diff
@@ -102,3 +109,121 @@ def match_by_attribute(attribute: str, content: str, patterns_database: Dict) ->
                     if matches:
                         libs.append({"containsLibrary": lib_name})
     return libs
+
+
+def download_database() -> Optional[Dict[str, Any]]:
```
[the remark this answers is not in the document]
willis89pr: Correct.

surfactant/infoextractors/native_lib_file.py
```diff
@@ -77,19 +81,19 @@ def extract_native_lib_info(filename):
...
         native_lib_info["nativeLibraries"].append({"containsLibrary": contains_library_names})
 
     return native_lib_info
 
 
-def match_by_attribute(attribute: str, content: str, patterns_database: Dict) -> List[Dict]:
-    libs = []
+def match_by_attribute(
+    attribute: str, content: Union[str, bytes], patterns_database: Dict[str, Any]
+) -> List[Dict[str, any]]:
```
nightlark, line 95:
```suggestion
) -> List[Dict[str, Any]]:
```

surfactant/infoextractors/native_lib_file.py
willis89pr: Woud it be:
`database[lib_name]["filecontent"].append(filecontent)`,

since, lib_name exists, and there may be file content parsed.

nightlark: My understanding of the `strict` type in the config file is that it means the filename must be a match, and then the "filecontent" regex is specifically for extracting the version number -- often with an extremely generic regex that matches basically anything that resembles a dotted version number. So I think we decided that we only want add the exact file name as a match for this case because the regex associated with `strict` patterns would result in way too many false positives.

willis89pr: In the else statement, sometimes the filename exists, and sometimes filecontent exists:
```
                if lib_name not in database:
                    database[lib_name] = {
                        "filename": [lib_name],
                        "filecontent": [],
                    }
                else:
                    print(database[lib_name])
```
output:
```
{'filename': ['bftpd'], 'filecontent': []}
{'filename': [], 'filecontent': ['gzip\\ [0-9](\\.[0-9]+)+?$', 'gzip\\ [0-9](\\.[0-9]+)+?$']}
{'filename': [], 'filecontent': ['gzip\\ [0-9](\\.[0-9]+)+?$', 'gzip\\ [0-9](\\.[0-9]+)+?$']}
{'filename': [], 'filecontent': ['^Inadyn\\ version\\ [0-9]+(\\.[0-9]+)+?\\ ']}
content regexp User-Agent: Siemens Canada Limited - ROX2 - [0-9](\.[0-9]+)?+$: multiple repeat at position 60
{'filename': ['radvd'], 'filecontent': []}
{'filename': ['radvdump'], 'filecontent': []}
{'filename': ['smbd'], 'filecontent': []}
{'filename': [], 'filecontent': ['NET-SNMP\\ version:\\ \\ [0-9](\\.[0-9]+)+?$', '^NET-SNMP\\ version:\\ [0-9](\\.[0-9]+)+?$']}
{'filename': ['snort'], 'filecontent': []}
{'filename': [], 'filecontent': ['traceroute\\.db\\:\\ Modern\\ traceroute\\ for\\ Linux,\\ version\\ [0-9](\\.[0-9]+)+?,\\ [[:alpha:]]{3}\\ [0-9]+\\ [0-9]+', '^Modern\\ traceroute\\ for\\ Linux,\\ [Vv]ersion\\ [0-9]+(\\.[0-9]+)+?']}
{'filename': [], 'filecontent': ['NET-SNMP\\ version:\\ \\ [0-9](\\.[0-9]+)+?$', '^NET-SNMP\\ version:\\ [0-9](\\.[0-9]+)+?$']}
{'filename': ['snort'], 'filecontent': []}
```

nightlark: That's because the entry already exists due to a previous line in the emba cfg file defining another pattern/filename for recognizing the same library. Each line/pattern is "independent" of the others.

I actually think the else case should check to make sure the same library name wasn't already added into the list of file names:

```python
if not lib_name in database[lib_name]["filename"]:
    database[lib_name]["filename"].append(lib_name)
```

The way the cfg specifies that a file name must be an exact match feels a bit like it was added on as an afterthought.

nightlark (approved):

PR #326, opened 2025-01-21
feat: Add configuration option to enable/disable Docker Scout

willis89pr:
- Integrated ConfigManager to retrieve 'enable_docker_scout' setting.
- Modified DockerScoutManager to respect the configuration option, allowing users to enable or disable Docker Scout functionality.
- Updated init_hook to conditionally check Docker Scout installation based on configuration.
- Added logic to short-circuit Docker Scout operations if disabled via configuration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,6 +118,14 @@ Getting the currently set value for the option would then be done with:
 surfactant config core.recorded_institution
 ```
 
+### Configuration Options
+
+- **docker.enable_docker_scout**: Controls whether Docker Scout is enabled. Default is `true`. To disable Docker Scout, run:
+  
+```bash
+surfactant config docker.enable_docker_scout false
+```
+
 ### Manual Editing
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -118,6 +118,14 @@ Getting the currently set value for the option would then be done with:
...
 ```
 
+### Configuration Options
+
+- **docker.enable_docker_scout**: Controls whether Docker Scout is enabled. Default is `true`. To disable Docker Scout, run:
+
+```bash
+surfactant config docker.enable_docker_scout false
+```
+
```
nightlark, line 128: Can you add an entry for it to https://github.com/LLNL/Surfactant/blob/main/docs/settings.md?

I don't think we need an extensive list of options in the README since we link to the settings page, but since this might be a common setting users want to disable -- maybe we can remove the heading (so it is part of the "Command Line" section and reword it as being "another example of a setting to control information gathered is..."

nightlark (approved):

PR #330, opened 2025-01-27
feat: Generate SBOM for a single file

### LINKED ISSUE #126: UX: Generate SBOM info for single file

nightlark:
Enable users to generate an SBOM for a single file, without having to create a config file. If a file name ends with `.json` it is treated as a specimen config file, rather than generating a single-file SBOM for the JSON file. To override this behavior, several prefixes can be prepended to the file path so the argument to Surfactant will be in the form `prefix:filepath` -- e.g. `file:abc.json`.

The prefixes recognized are:
* `file` to force Surfactant to generate an SBOM from the single file given after the prefix. Likely only needed to generate a single file SBOM for a JSON file or a file whose name starts with one of the special prefixes as part of the file name, both of which should be rare. e.g. `file:config:abc.xyz` would generate a single file SBOM for a file with the (odd) name `config:abc.xyz`
* `config` to force Surfactant to treat the given file path as a Surfactant specimen config file. This should never be needed, unless a user decides to omit a `.json` file extension for their Surfactant specimen config file names.
* `dir` to force Surfactant to treat the given file path as a specific directory to generate an SBOM for, with "implied" specimen config file settings (summarized below). This should only be needed if a directory has one of these special prefixes at the start of its name, which is very rare. e.g. `dir:config:mydirectory` would generate an SBOM over the contents in a directory named `config:mydirectory`.

For generating a single file or single directory SBOM some assumptions are made for parameters that a specimen config file would usually contain:
* `extractPaths`: the path given as an argument to Surfactant
* `installPrefix`: either the directory given as the argument to Surfactant, or the parent directory if the path given to Surfactant is for a single file SBOM; a relative path to a file will result in this being a relative path while an absolute path will result in this being the absolute path, which gives some control over the install paths present in the resulting SBOM

Resolves #126

### DIFF
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -76,12 +76,35 @@ pip install -e ".[test,dev]"
 `pip install` with the `-e` or `--editable` option can also be used to install Surfactant plugins for development.
 
 ## Generating an SBOM
-
-To create an SBOM, run the `surfactant generate` subcommand. For more details on the options it takes, please refer to this page on [basic usage](basic_usage.md). For more information on writing Surfactant configuration files for software specimens, see the [configuration files](configuration_files.md) page.
+To create an SBOM, run the `surfactant generate` subcommand. For more details on the options it takes, please refer to this page on [basic usage](basic_usage.md). For more information on writing Surfactant configuration files for software specimens, see the documentation on how to build a [specimen configuration file](configuration_files.md#specimen-configuration-file).
 
 The following diagram gives a high-level overview of what Surfactant does. The [internal implementation overview](internals_overview.md) page gives more detail about how Surfactant works internally.
 
 ![Surfactant Overview Diagram](img/surfactant_overview_diagram.svg)
+
+In simpler cases such as generating an SBOM for a single file or directory that lives on the same system as Surfactant is being run on, Surfactant can just be given the path to generate the SBOM for:
+
+```bash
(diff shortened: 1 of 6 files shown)


docs/getting_started.md
```diff
@@ -76,13 +76,36 @@ pip install -e ".[test,dev]"
...
+
+Similarly, a `config:` prefix forces Surfactant to interpret the given file path as a specimen configuration file regardless of if the file name is missing a `.json` extension.
+
+A file or directory name that starts with one of these special prefixes could cause problems, however these cases should be extremely rare and can always be solved by creating a specimen configuration file (which since it is user created, can be given a file name that avoid s issues). However, a special prefix could also be used to solve the issue. For example with a directory named `config:myapp`, running `surfactant generate config:myapp` will look for a specimen configuration file called `myapp`. To resolve this, the `dir:` prefix could be added to essentially tell Surfactant "this directory is actually named config:myapp". Running `surfactant generate dir:config:myapp` would then generate an SBOM for everyting in a directory called `config:myapp`.
```
willis89pr, line 103: Typo: "...that avoid s issues..."

willis89pr (approved):

PR #332, opened 2025-02-02
Port web-based UI to a Textual TUI

KendallHarterAtWork:
### Summary

If merged this pull request will add a Textual based TUI with the aim to replace the web-based UI

### Proposed changes

This is purely adding a new command `surfactant tui` that allows functionality equal to the web-based UI but with easier file selection.  It also fixes a bug in toggling the dark/light mode in the config_tui.


surfactant/cmd/tui.py
```diff
@@ -0,0 +1,484 @@
...
+# TODO: Rewrite to use ContentSwitcher?
+class TUI(textual.app.App):
+    """An app for running Surfactant commands"""
+
+    BINDINGS = [
+        ("d", "toggle_dark", "Toggle dark mode"),
+        ("q", "quit", "Quit"),
+    ]
+    TITLE = "Surfactant TUI"
+    CSS_PATH = "../web-files/tui.tcss"
```
nightlark, line 437: Can we move this to a different folder that isn't web-specific? Maybe along the lines of "ui-resources" or "ui-data"?

nightlark:
Overall usage-wise it looks good, and easier to work with than copying paths into a web browser. The main quirk I noticed is the file selection -- it's a bit strange how the output directory field is used to select both the directory and the file name (that then gets split across two different text fields). It feels like calling the "button" select file would be clearer, but I guess that doesn't allow a user to type in a name for a new file to create.

The other related quirk is when selecting a directory instead of a file, it sticks the selected subfolder name as the file name.
e.g. selecting `HOME/Surfactant/plugins` in the file browser, the directory field gets set to `HOME/Surfactant` and the file name is set to `plugins` (which is a folder rather than a file).

nightlark (approved):

PR #338, opened 2025-02-04
Multiple json db

wangmot:
### Summary

If merged this pull request will allow multiple json databases to be used for pattern matching. Users can use private databases--they will just have to be in the same format as the emba.json and they will have to be placed in the correct location.

### Proposed changes

I've created a new dictionary of dictionaries to store all the json databases (each json file as a single entry in the dictionary).

### DIFF
--- a/surfactant/infoextractors/native_lib_file.py
+++ b/surfactant/infoextractors/native_lib_file.py
@@ -221,6 +221,6 @@ def init_hook(command_name: Optional[str] = None) -> None:
         native_lib_manager.load_db()
         logger.info("Initializing native_lib_file complete.")
 
-        # Create native_lib_patterns folder for storing JSON DB's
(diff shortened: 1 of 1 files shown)


wangmot:
I am noticing one bug. Let's take the `busybox` example binary which is in the `emba.json` database. When the patterns get matched against the `emba.json`, it finds 3 matches in the `containsLibrary` list. But when I have two json databases, both get added to a new dictionary and if I remove the `busybox` entry from `emba.json` and add it to the new json database, all of a sudden when the pattern matching occurs, it only returns one match. It seems like it breaks out of the loop once a single match is found. But this behavior doesn't occur with the original `emba.json` database. With that one, it is capable of finding multiple matches and only breaks out once all the patterns are traversed.

I'm guessing the issue is with the way the two databases are traversed and added to the new dictionary, something weird is happening. Currently looking into this now.

PR #347, opened 2025-02-11
fix: Specimen config includeAllFiles option being ignored

### LINKED ISSUE #342: Fix includeAllFiles specimen config file entry
includeAllFiles in a specimen config file doesn't work. `entry.includeAllFiles` needs to be added to this if statement: https://github.com/LLNL/Surfactant/blob/main/surfactant/cmd/generate.py#L468

nightlark:
Resolves #342

Add missing condition to include a check for includeAllFiles option from specimen configuration files.


wangmot (approved):
Tested with a sample file with unrecognized format, `test.xyz` and tested with `includeAllFiles` set to true and tested it without that option in the config file. 

`includeAllFiles` set to true shows it in the resulting SBOM!

PR #348, opened 2025-02-11
Rename include all files to omit unrecognized types

nightlark:
Changes the default Surfactant behavior to include all files by default. This required changing defaults for both the specimen config file entry and the command line argument fallback to true, and the logic from `or` to `and` so that if the user sets either option to false, the file will be excluded. This change was needed for the config file `includeAllFiles` option to work, if `or` was used the default value of the command line option being true would essentially override it.


wangmot (approved):
Tested with a sample file with non-standard file extension, `test.xyz`, and resulting SBOM recognized this file.

wangmot:
Update: 

Having `"omitUnrecognizedTypes": true` in config file does `NOT` show sample `test.xyz` file in resulting SBOM.
 
`surfactant generate Helics-3.5.json output` DOES show the `test.xyz` file. (This should happen since it's the default).
 
`surfactant generate Helics-3.5.json output --omit_unrecognized_types` does NOT show the `test.xyz` file in the resulting SBOM. 

Changing `includeAllFiles` to `--omit_unrecognized_types` makes this more intuitive.

wangmot (approved):

PR #353, opened 2025-02-21
CI Workflow for 3rd-Party Plugins

willis89pr:
### Summary

GH CI workflow for third party plugins and example grype pytest.

If merged this pull request will create a GH CI workflow for 3rd party plugins and implements an example pytest for the grype plugin. The CI workflow iterates through the plugin directory look for test/ directory and runs the pytests if one exists in the test/ directory.

### Proposed changes

GH CI Workflow.
grype pytest.


.github/workflows/test_plugins.yml
```diff
@@ -0,0 +1,98 @@
...
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      # Step 3: Enable Debug Logging
+      - name: Enable Debug Logging
+        run: echo "Enabling debug logging..."
+        env:
+          ACTIONS_STEP_DEBUG: true
```
nightlark, line 36: This is probably okay to remove now.

plugins/grype/surfactantplugin_grype.py
```diff
@@ -4,34 +4,55 @@
 # SPDX-License-Identifier: MIT
 import gzip
 import subprocess
+
+# Configure loguru to show logs at DEBUG level or higher
```
nightlark, line 8: This feels like it should be attached to line 23

plugins/grype/surfactantplugin_grype.py
```diff
@@ -4,34 +4,55 @@
...
 
 import surfactant.plugin
+from surfactant.plugin.manager import (  # Import the helper function
+    get_plugin_manager,
+    is_plugin_blocked,
+)
 from surfactant.sbomtypes import SBOM, Software
 
+logger.remove()  # Remove the default logger configuration
+logger.add(sys.stdout, level="DEBUG")  # Add a new logger that outputs to console
```
nightlark, line 23: Does this mess up the logger used in the rest of Surfactant? https://github.com/LLNL/Surfactant/blob/main/surfactant/__main__.py#L55-L5

Instead of changing the logger settings here, in the tests when you run Surfactant you can pass it the --log-level option.

plugins/grype/surfactantplugin_grype.py
```diff
@@ -69,24 +90,51 @@ def run_grype(filename: str) -> object:
 def extract_file_info(
     sbom: SBOM, software: Software, filename: str, filetype: str, children: list
 ) -> Optional[List[Software]]:
-    if disable_plugin or filetype not in ("DOCKER_TAR", "DOCKER_GZIP"):
+    logger.info(f"Extracting file info for {filename} of type {filetype}")
```
nightlark, line 93: I think this log message can be removed. Will make the log output quite verbose since it will be printed for every file.

plugins/grype/surfactantplugin_grype.py
```diff
@@ -69,24 +90,51 @@ def run_grype(filename: str) -> object:
...
+
+    if disable_plugin:
+        logger.warning("Grype plugin is disabled. Skipping file analysis.")
```
nightlark, line 96: Same with this message -- I'd expect it to get printed for every single file when grype is disabled.

plugins/grype/surfactantplugin_grype.py
```diff
@@ -69,24 +90,51 @@ def run_grype(filename: str) -> object:
...
+        return None
+
+    if filetype not in ("DOCKER_TAR", "DOCKER_GZIP"):
+        logger.debug(f"File type {filetype} is not supported by the grype plugin.")
```
nightlark, line 100: Ditto here -- most files will not be ones that the grype plugin can handle. Maybe for the "trace" log level (I'd also lean towards bumping up the log level for the next two debug messages as well).

surfactant/plugin/manager.py
```diff
@@ -96,6 +96,25 @@ def set_blocked_plugins(pm: pluggy.PluginManager) -> None:
...
+        pm (pluggy.PluginManager): The plugin manager instance.
+        plugin_name (str): The name of the plugin to check.
+
+    Returns:
+        bool: True if the plugin is blocked, False otherwise.
+    """
+    if pm.is_blocked(plugin_name):
+        logger.warning(
+            f"The plugin '{plugin_name}' is blocked. To unblock run, 'surfactant plugin enable {plugin_name}'"
+        )
```
nightlark, line 113: I'd lean towards letting whatever code is calling this function decide whether or not it needs to print a warning message.

plugins/grype/surfactantplugin_grype.py
```diff
@@ -4,34 +4,55 @@
...
-        logger.warning("Install grype for the grype plugin to run")
-    return result == 0
+        logger.error("Grype is not installed. Please install it to use this plugin.")
+        return False
 
 
-disable_plugin = not check_if_grype_installed()
+# Initialize the plugin manager
+plugin_manager = get_plugin_manager()
+disable_plugin = not check_if_grype_installed() or is_plugin_blocked(plugin_manager, __name__)
```
nightlark, line 41:
```suggestion
disable_plugin = not check_if_grype_installed() or is_plugin_blocked(plugin_manager, __name__)
```

If the plugin is blocked the extract_file_info hook should never get called, so  the value of`disable_plugin` will never be checked in the case that it is set to True due to the is_plugin_blocked call.

plugins/grype/surfactantplugin_grype.py
willis89pr: I changed this to debug****

willis89pr (commented):
Made changes.

plugins/grype/surfactantplugin_grype.py
willis89pr: Should logger.debug be removed here?

nightlark: This one should either be removed or changed to "trace". I'd lean towards remove, since they will both create a ton of noise in the output log (e.g. if every plugin is changed to log messages in this scenario, I'd expect at least 10 lines of output that are irrelevant per file that will bury the messages that are likely to be more relevant to debugging issues).



```
File type PE is not supported by the grype plugin.
File type PE is not supported by the syft plugin.
File type PE is not supported by the elf plugin.
File type PE is not supported by the docker plugin.
File type PE is not supported by the  plugin.
```

plugins/grype/requirements-test.txt
```diff
@@ -0,0 +1,4 @@
+# Test dependencies for the Grype plugin
+docker>=6.1.3
+requests>=2.31.0
+pytest>=7.0.0
```
nightlark, line 4: These dependencies can be added to the pyproject.toml for the plugin under an optional dependency group instead. Then in the workflow, pip install them along with the plugin by specifying the optional dependencies to install in square brackets (e.g. `pip install -e .[test]` to install all of the `test` optional dependencies).

plugins/grype/requirements-test.txt
nightlark, line 1: This file can be deleted now.

nightlark (approved):

PR #360, opened 2025-03-04
File decompression

wangmot:
### Summary

If merged this pull request will decompress compressed file formats like `zip`, `tar`, `tar.gz`, `tar.bz2`, and `tar.xz` and then add them to the queue for surfactant to traverse. 

### Proposed changes

This file decompression support is implemented as a new plugin. To make it work, the `ContextEntry` class had to be updated to accept one more parameter called `skipProcessingArchive: Optional[bool] = False`. 

The issue was that the parent archive was repeatedly decompressed and then put back on the queue, causing an infinite loop.


surfactant/infoextractors/file_decompression.py
```diff
@@ -0,0 +1,110 @@
...
+    compression_formats = {
+        ".zip": "zip",
+        ".tar": "tar",
+        ".tar.gz": "tar.gz",
+        ".tar.bz2": "tar.bz2",
+        ".tar.xz": "tar.xz",
+    }
+    for ext, format in compression_formats.items():
+        if filename.endswith(ext):
+            return format
```
nightlark, line 33:
```suggestion
    for ext, fmt in compression_formats.items():
        if filename.endswith(ext):
            return fmt
```

Fix pylint: `surfactant/infoextractors/file_decompression.py:31:13: W0622: Redefining built-in 'format' (redefined-builtin)`

nightlark, line 33: Actually, you should use the filetype argument that extract_file_info gets passed -- and just match against the corresponding strings defined in https://github.com/LLNL/Surfactant/blob/main/surfactant/filetypeid/id_magic.py

Doing it by filename means someone could lie and name their amazing new custom file "run.zip". Or they might decide to rename their "abc.tar.gz" file to just "abc". In either case, file extension as a check will fail. The filetype argument is from id_magic which uses the actual file contents to detect the file type.

surfactant/infoextractors/file_decompression.py
```diff
@@ -0,0 +1,110 @@
...
+def create_temp_dir():
+    # Create a temporary directory
+    temp_dir = tempfile.mkdtemp(prefix="surfactant-temp")
+    return temp_dir
+
+
+def decompress_zip_file(filename):
+    temp_folder = create_temp_dir()
+    with zipfile.ZipFile(filename, "r") as zip:
+        zip.extractall(path=temp_folder)
```
nightlark, line 88:
```suggestion
    with zipfile.ZipFile(filename, "r") as f:
        f.extractall(path=temp_folder)
```

Fix pylint warning: `surfactant/infoextractors/file_decompression.py:87:43: W0622: Redefining built-in 'zip' (redefined-builtin)`

surfactant/infoextractors/file_decompression.py
```diff
@@ -0,0 +1,110 @@
...
+    if compression_format == "zip":
+        temp_folder = decompress_zip_file(filename)
+    elif compression_format == "tar":
+        temp_folder = extract_tar_file(filename)
+    elif compression_format in {"tar.gz", "tar.bz2", "tar.xz"}:
+        tar_modes = {
+            "tar.gz": "r:gz",
+            "tar.bz2": "r:bz2",
+            "tar.xz": "r:xz",
+        }
```
nightlark, line 71: The compression_format checked against should be the `filetype` argument strings (possible values defined in id_magic.py)

surfactant/infoextractors/file_decompression.py
```diff
@@ -0,0 +1,110 @@
...
+    temp_folder = check_compression_type(filename, compression_format)
+
+    # Add a new ContextEntry for the temp dir
+    new_entry = ContextEntry(
+        archive=filename, installPrefix="", extractPaths=[temp_folder], skipProcessingArchive=True
+    )
+
+    # Add new ContextEntry to queue
+    context.put(new_entry)
+    logger.info(f"New ContextEntry added for extracted files: {temp_folder}")
```
nightlark, line 56:
```suggestion
    logger.info(f"New ContextEntry added for extracted files: {temp_folder}")
    return None
```

Fix pylint warning: `surfactant/infoextractors/file_decompression.py:38:0: R1710: Either all return statements in a function should return an expression, or none of them should. (inconsistent-return-statements)`

surfactant/infoextractors/file_decompression.py
```diff
@@ -0,0 +1,111 @@
...
+    }
+    for ext, fmt in compression_formats.items():
+        if filename.endswith(ext):
+            return fmt
+    return None
+
+
+@surfactant.plugin.hookimpl
+def extract_file_info(
+    sbom: SBOM, software: Software, filename: str, filetype: str, context: Queue[ContextEntry]
```
nightlark, line 39:
```suggestion
    sbom: SBOM, software: Software, filename: str, filetype: str, context: "Queue[ContextEntry]"
```

I think quoting the type will fix the Python 3.8 TypeError in the CI tests.

surfactant/infoextractors/file_decompression.py
wangmot: Haha "their amazing new custom file"

surfactant/infoextractors/file_decompression.py
wangmot: Yaaassss, that did it!

surfactant/filetypeid/id_magic.py
```diff
@@ -201,6 +201,12 @@ def identify_file_type(filepath: str) -> Optional[str]:
                 if cm == 8:
                     if (cmf * 256 + flg) % 31 == 0:
                         return "ZLIB"
+
+            # Check for compressed TAR files
+            if magic_bytes[:3] == b"BZh":
+                return "TAR BZIP2"
```
nightlark, line 207:
```suggestion
                return "BZIP2"
```

surfactant/filetypeid/id_magic.py
```diff
@@ -201,6 +201,12 @@ def identify_file_type(filepath: str) -> Optional[str]:
...
+            if magic_bytes[:6] == b"\xfd\x37\x7a\x58\x5a\x00":
+                return "TAR XZ"
```
nightlark, line 209:
```suggestion
                return "XZ"
```

surfactant/filetypeid/id_magic.py
```diff
@@ -201,6 +201,12 @@ def identify_file_type(filepath: str) -> Optional[str]:
                 if cm == 8:
                     if (cmf * 256 + flg) % 31 == 0:
                         return "ZLIB"
+
+            # Check for compressed TAR files
```
nightlark, line 205: Often, but not always. Compressed tar files are like onions, they have layers (compression algorithm, and decompression results in a file with a type that usually happens to be a TAR file).

nightlark: These two checks could take place farther up in this file, around where the check for GZIP is.

nightlark (approved):

PR #376, opened 2025-03-18
add current_context variable to extract_file_info hook

wangmot:
### Summary

If merged this pull request will add an additional optional parameter called `current_context`. Plugins can now access the current entry in the queue. 

### Proposed changes

Currently, we have the `context` parameter, which references the context queue. This `context` param has been renamed to `context_queue` to make it more clear what it is.

### DIFF
--- a/surfactant/cmd/generate.py
+++ b/surfactant/cmd/generate.py
@@ -30,6 +30,7 @@ def real_path_to_install_path(root_path: str, install_path: str, filepath: str) 
 
 def get_software_entry(
     context_queue,
+    current_context,
     pluginmanager,
     parent_sbom: SBOM,
     filepath,
@@ -64,7 +65,7 @@ def get_software_entry(
             filename=filepath,
             filetype=filetype,
             context_queue=context_queue,
-            current_context=None,
+            current_context=current_context,
             children=sw_children,
             software_field_hints=sw_field_hints,
             omit_unrecognized_types=omit_unrecognized_types,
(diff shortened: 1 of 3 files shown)


surfactant/plugin/hookspecs.py
```diff
@@ -50,8 +51,9 @@ def extract_file_info(
         software (Software): The software entry the gathered information will be added to.
         filename (str): The full path to the file to extract information from.
         filetype (str): File type information based on magic bytes.
-        context (Queue[ContextEntry]): Modifiable queue of entries from input config file. Existing plugins should
+        context_queue (Queue[ContextEntry]): Modifiable queue of entries from input config file. Existing plugins should
```
nightlark, line 54:
```suggestion
        context_queue (Queue[ContextEntry]): Modifiable queue of entries typically initialized from the input specimen config file. Plugins can add new entries to the queue to make Surfactant process additional files/folders. Existing plugins should
```

surfactant/cmd/generate.py
```diff
@@ -29,7 +29,7 @@ def real_path_to_install_path(root_path: str, install_path: str, filepath: str)
 
 
 def get_software_entry(
-    context,
```
nightlark, line 32: current_context will need to be added as an argument for this function, then passed along to the hook call instead of None.

surfactant/cmd/generate.py
```diff
@@ -303,15 +304,16 @@ def sbom(
         file_symlinks: Dict[str, List[str]] = {}
         # List of filename symlinks; keys are SHA256 hashes, values are file names
         filename_symlinks: Dict[str, List[str]] = {}
-        while not context.empty():
-            entry: ContextEntry = context.get()
+        while not context_queue.empty():
+            entry: ContextEntry = context_queue.get()
+            current_context = entry
```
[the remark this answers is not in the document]
wangmot: Okay I see what you're saying. I had to pass None in earlier cause it wasn't recognizing `entry`, because I wasn't passing `current_entry` to `get_software_entry`. I left the param as it was: `entry`.

nightlark (approved):

Issue #366, opened 2025-03-12
Update file decompression to recognize if archive file needs extraction


nightlark:
Currently, the "archive" field in specimen configuration files/context entries has a type that is not identified, so extraction doesn't happen. #365 will make so that its type gets identified and an attempt at extracting the file contents will occur -- however, if the archive contents are already extracted (manually or otherwise indicated by extractPaths being not empty) then extraction should not happen.

Summary: in the file decompression info extractor, if the current context has an archive but empty list of extract paths, then add a context entry to the queue for the archive, and let the context entry added to the queue inherit the install prefix from the current context (depends on #364; changes made in #365 will work better after this change is made).

Issue #365, opened 2025-03-12
Do file type identification for "archive" file


nightlark:
Update file type identification to run on “archive” file, to pass to extract file info hook call for the archive (change to be made in generate.py somewhere)

PR #379, opened 2025-03-31
add context entry for archives

wangmot:
### Summary

If merged this pull request will allow archives listed in the config file to be extracted. This depends on issue https://github.com/LLNL/Surfactant/issues/365 to be resolved. 365 will make sure that the archive file has an appropriate `filetype`.

Here is a sample config file:

Screenshot 2025-03-30 at 11 24 29 PM

Using the sample config file above, if we have an `archive` key and an empty list of `extractPaths`, no extraction occurs. With this PR, we can add a context entry to the queue for the archive and have the context entry added to the queue inherit the install prefix from the current context. Linked to issue https://github.com/LLNL/Surfactant/issues/366

### Proposed changes

Add a new context entry if the current context has an archive but empty list of extract paths.

### DIFF
--- a/surfactant/infoextractors/file_decompression.py
+++ b/surfactant/infoextractors/file_decompression.py
@@ -56,30 +56,28 @@ def extract_file_info(
 
     # Check that archive key exists and filename is same as archive file
     if current_context.archive and current_context.archive == filename:
-        # If archive has empty list of extractPaths, an entry to the queue will be added
-        if current_context.extractPaths == []:
-            temp_folder = check_compression_type(current_context.archive, compression_format)
-            install_prefix = current_context.installPrefix
-            extract_paths = [temp_folder]
-            logger.info(f"New ContextEntry added for archive: {current_context.archive}")
-        # Assume archive is already extracted if the extractPaths is not empty
-        else:
+        if current_context.extractPaths is not None and current_context.extractPaths != []:
             logger.info(
                 f"Already extracted, skipping extraction for archive: {current_context.archive}"
             )
             return None
-    # Assume there is a compressed file in the extractPaths
-    elif current_context.installPrefix == "/":
-        # Decompress the file based on its format
-        temp_folder = check_compression_type(filename, compression_format)
-        extract_paths = [temp_folder]
-        logger.info(f"New ContextEntry added for extracted files: {temp_folder}")
+
+        # Inherit the context entry install prefix for the extracted files
+        install_prefix = current_context.installPrefix
+
+    # Decompress the file based on its format
+    temp_folder = check_compression_type(filename, compression_format)
+    extract_paths = [temp_folder]
 
     # Create a new context entry and add it to the queue
     new_entry = ContextEntry(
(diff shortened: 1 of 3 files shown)


surfactant/infoextractors/file_decompression.py
```diff
@@ -46,21 +46,40 @@ def extract_file_info(
...
+            temp_folder = check_compression_type(current_context.archive, compression_format)
+            install_prefix = current_context.installPrefix
+            extract_paths = [temp_folder]
+            logger.info(f"New ContextEntry added for archive: {current_context.archive}")
+        # Assume archive is already extracted if the extractPaths is not empty
+        else:
+            logger.info(
+                f"Already extracted, skipping extraction for archive: {current_context.archive}"
+            )
+            return None
```
nightlark, line 70:
```suggestion
        # Assume archive is already extracted if the extractPaths is not empty and skip it
        # Otherwise, inherit the context entry install prefix for the extracted files
        if current_context.extractPaths != []:
            logger.info(
                f"Already extracted, skipping extraction for archive: {current_context.archive}"
            )
            return None
        else:
            install_prefix = current_context.installPrefix
```

This can be simplified so only a single call to decompress later on is needed

surfactant/infoextractors/file_decompression.py
```diff
@@ -46,21 +46,40 @@ def extract_file_info(
...
+    # Assume there is a compressed file in the extractPaths
+    elif current_context.installPrefix == "/":
+        # Decompress the file based on its format
+        temp_folder = check_compression_type(filename, compression_format)
+        extract_paths = [temp_folder]
+        logger.info(f"New ContextEntry added for extracted files: {temp_folder}")
```
nightlark, line 76:
```suggestion

    # Decompress the file based on its format
    temp_folder = check_compression_type(filename, compression_format)
    extract_paths = [temp_folder]
    logger.info(f"New ContextEntry added for extracted files: {temp_folder}")
```

Only a single call to decompress a file is needed, there shouldn't be anything special to this part that makes decompressing a file listed as an "archive" different than one found within the extract paths folder that surfactant is walking through.

surfactant/infoextractors/file_decompression.py
```diff
@@ -46,21 +46,40 @@ def extract_file_info(
...
+
+    # Create a new context entry and add it to the queue
     new_entry = ContextEntry(
-        archive=filename, installPrefix="", extractPaths=[temp_folder], skipProcessingArchive=True
+        archive=filename, installPrefix="", extractPaths=extract_paths, skipProcessingArchive=True
```
nightlark, line 80:
```suggestion
        archive=filename, installPrefix=install_prefix, extractPaths=extract_paths, skipProcessingArchive=True
```

If the file being processed is not the current_context.archive, then install_prefix has its default value of "" -- which is fine for files found within extract paths folders that are being walked.

If the file being processed is the current_context.archive (which is the same as filename), then install_prefix will have been set to inherit the install prefix set in current_context (or this line will never be reached if current_context.extractPaths is not empty and a return statement was already reached).

nightlark (approved):

PR #384, opened 2025-04-07
Container Prefix support

KendallHarterAtWork:
### Summary

If merged this pull request will address #130. 

### Proposed changes

This adds a `containerPrefix` to the Surfactant configuration files, allowing subdirectories of archives to have properly container paths.

### DIFF
--- a/surfactant/cmd/generate.py
+++ b/surfactant/cmd/generate.py
@@ -42,14 +42,14 @@ def get_software_entry(
     user_institution_name="",
     omit_unrecognized_types=False,
     skip_extraction=False,
-    container_prefix=None
+    container_prefix=None,
 ) -> Tuple[Software, List[Software]]:
     sw_entry = Software.create_software_from_file(filepath)
     if root_path is not None and install_path is not None:
         sw_entry.installPath = [real_path_to_install_path(root_path, install_path, filepath)]
     if root_path is not None and container_uuid is not None:
         # Setup the container prefix as needed
-        prefix = container_prefix if container_prefix is not None else ''
+        prefix = container_prefix if container_prefix is not None else ""
         if prefix != "":
             prefix = "/" + prefix
             # Remove slash at the end since it results in incorrect paths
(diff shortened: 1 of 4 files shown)


surfactant/context.py
```diff
@@ -27,6 +28,8 @@ class ContextEntry:
             will still be included.
         skipProcessingArchive (Optional[bool]): If True, skip processing the given archive file with info extractors.
             Software entry for the archive file will only contain basic information such as hashes. Default is False.
+        containerPrefix (Optional[str]): The prefix to use for the generated SBOM's containerPath.  To be used when
+            `extractPaths` points to a subfolder of an archive.
```
nightlark, line 32:
```suggestion
        containerPrefix (Optional[str]): The prefix to use for the generated SBOM's containerPath.  Used to indicate that the
            `extractPaths` specified should map to a specific subfolder within the corresponding archive file.
```

surfactant/cmd/generate.py
```diff
@@ -42,16 +42,26 @@ def get_software_entry(
...
     if root_path is not None and install_path is not None:
         sw_entry.installPath = [real_path_to_install_path(root_path, install_path, filepath)]
     if root_path is not None and container_uuid is not None:
+        # Setup the container prefix as needed
+        prefix = container_prefix if container_prefix is not None else ""
+        if prefix != "":
+            prefix = "/" + prefix
+            # Remove slash at the end since it results in incorrect paths
+            if prefix.endswith("/"):
+                prefix = prefix[:-1]
```
nightlark, line 57:
```suggestion
        prefix = container_prefix.strip("/") if container_prefix is not None else ""
        if prefix != "":
            prefix = "/" + prefix
```

Strip forward slashes from the start and end of the containerPrefix -- will make sure we don't have excessive slashes at the start or end. Maybe could also make sure any use of Windows style backslash directory separators are replaced by UNIX style forward slashes.

This entire transformation block could be moved to the start of the sbom() function so that it only happens once per context entry rather than potentially thousands of times for each file within an extract path -- around line 343 where a similar transformation and checks are done for the installPrefix.

nightlark (approved):

PR #388, opened 2025-04-16
feat: Add fast pattern matching using AhoCorasick searching

### LINKED ISSUE #306: Fast pattern matching
Regex is nice due to its ubiquity (and Python having a built-in library for it). But it is slow to match hundreds of patterns against large files. An algorithm such as Aho-corasick that can simultaneously match multiple patterns with an FSM and only needs to go through a file approx. once should be much faster.

Look into options for recognizing the static string prefix portion of a regular expression (that hopefully is not an empty string), and turn all of the prefixes identified into something that can do a fast first pass match, and only check the regular expressions that match the prefix once a match for the prefix is found in a file.

hyperscan via the pyperscan (https://pypi.org/project/pyperscan) PyPI package is another possibility (would need to fall back on the existing code built-in `re` module, but on platforms where it is available this might be close to a drop-in replacement).

These two options may not be mutually exclusive either -- if both get implemented and hyperscan is significantly faster, then we could use hyperscan when available, then fall-back to Aho-Corasick pre-filtering as a portable pure Python option.

nightlark:
### Summary

If merged this pull request will add support for quickly locating places in a file that might match a regular expression. It results in the runtime remaining roughly constant as the number of patterns increases, enabling matching of 100k+ patterns in under a minute in situations that would take over 30 minutes when checking each regex individually. For larger files, it will result in the time taken being 1/n of the current time, where `n` is the number of patterns being searched for.

### Proposed changes

* Add fixed regex literal extraction code, by parsing regex into a tree and searching for literals with some support branches/subpatterns/character classes that add multiple possibilities up to a certain threshold
* Add aho-corasick automaton builder to quickly find "needles in a haystack" that match the identified regex literals
* Use aho-corasick automaton in native lib detection to quickly search file contents for places that are most likely to match the regular expression, and then check if the regular expression matches the contents in the part of the file that was identified

A minimum prefix length of 3 is used to avoid an excessive number of matches against extremely common characters (such as 0-9), and the maximum number of prefixes computed per pattern is capped at 10 to avoid needing to construct an extremely large automaton (changing this value is a balance between higher likelihood of matching the regex at identified locations vs needing to check more places for a regex match -- 5-15 seemed like a bit of a sweet spot).

Resolves #306 - performance improvement seems large enough that hyperscan likely won't yield a similar order of magnitude speedup over this approach. Could still explore that option in the future, but this is looking pretty good for now.


nightlark:
Reworked things a bit to fix a corner case where the regex literal extractor was too eager to continue adding more to prefixes, even if the existing set of prefixes didn't cover all cases. The one pattern this really affected was for nlohmann:

* Pattern: `nlohmann[0-9]+json_abi_v`
* Previously: prefixes = ["nlohmann0json_abi_v", "nlohmann1json_abi_v", etc]
* Now: prefixes = ["nlohmann0", "nlohmann1", etc]

Essentially, the issue is that the regex has `[0-9]+`, which can expand to an infinite number of possible prefixes -- so continuing to add characters from later in the regex to build up the prefix will miss all of those cases with extra numbers.

wangmot (approved):

wangmot:
Tested with a sample regex pattern database with ~15,000 patterns. Ago-corasick matching took about 7 seconds and current matching took about 5 mins 10 seconds. Ran on a directory of size 18.9 MB.

PR #392, opened 2025-04-25
TUI improvements

KendallHarterAtWork:
### Summary

If merged this pull request will add some improvements to the TUI, mainly by switching the main content to use TabbedContent, which should fix some minor display issues and be more reactive, and adding containerPrefix to the TUI.


nightlark:
It looks like most of the directory selection UI elements are working better now, though the one for picking a `Config directory` under the `Config` tab doesn't have the `Select directory` button.

nightlark (approved):

PR #405, opened 2025-05-13
Gzip garbage bytes

wangmot:
### Summary

If merged this pull request will stop surfactant from crashing if it encounters a GZIP file with trailing garbage bytes. 

### Proposed changes

Added `gzip.BadGzipFile` and `OSError` exceptions. `gzip.BadGzipFile` specifically works with GZIP files and `OSError` is caught when the program comes across corrupted BZIP2 and XZ files.

### DIFF
--- a/surfactant/infoextractors/file_decompression.py
+++ b/surfactant/infoextractors/file_decompression.py
@@ -69,15 +69,16 @@ def extract_file_info(
     temp_folder = check_compression_type(filename, compression_format)
     extract_paths = [temp_folder]
 
-    # Create a new context entry and add it to the queue
-    new_entry = ContextEntry(
-        archive=filename,
-        installPrefix=install_prefix,
-        extractPaths=extract_paths,
-        skipProcessingArchive=True,
-    )
-    context_queue.put(new_entry)
-    logger.info(f"New ContextEntry added for extracted files: {temp_folder}")
+    if temp_folder is not None:
+        # Create a new context entry and add it to the queue
+        new_entry = ContextEntry(
+            archive=filename,
+            installPrefix=install_prefix,
+            extractPaths=extract_paths,
+            skipProcessingArchive=True,
+        )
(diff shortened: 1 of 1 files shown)


surfactant/infoextractors/file_decompression.py
```diff
@@ -140,18 +140,21 @@ def decompress_file(filename, compression_type):
...
+        elif compression_type == "BZIP2":
+            with bz2.open(filename, "rb") as f_in:
+                with open(os.path.join(temp_folder, output_filename), "wb") as f_out:
+                    shutil.copyfileobj(f_in, f_out)
+        elif compression_type == "XZ":
+            with lzma.open(filename, "rb") as f_in:
+                with open(os.path.join(temp_folder, output_filename), "wb") as f_out:
+                    shutil.copyfileobj(f_in, f_out)
+    except (gzip.BadGzipFile, OSError) as e:
+        print(f"Warning: Trailing garbage bytes or concatenated streams ignored {filename}: {e}")
```
nightlark, line 157: This should be a logger warning instead of print (message can also be updated to not say "warning"; perhaps change it to be more generic such "unable to uncompress {filename}: {e}" and let the error `e` provide details on why since for bzip2 and xz it sounded like it is corruption rather than trailing garbage/concatenated streams that cause the error).

wangmot: Okay, good point. I went ahead and separated the two exceptions: `gzip.BadGzipFile` and `OSError` because if I return `None` for both, then we wouldn't be able to use the part that was already decompressed before the trailing garbage bytes cause the crash for `gzip.BadGzipFile`. 

With the new implementation, we can keep what was already decompressed and still return the `temp_folder` to later be added to the context_queue before the garbage bytes at the end crash surfactant and make none of it usable.

nightlark (approved):

PR #450, opened 2025-07-11
Use extractcode as stopgap for decompression/extraction

WorkingRobot:
### Summary

If merged this pull request will add support for a plethora of compression types and archive formats as a best-effort stopgap until we decide to add support for these formats manually (if necessary) to add metadata or version info (e.g. the RPM files that @matthewkelley22 is working on).

### Proposed changes

Removes the .rar file decompression things and replaces it entirely with `extractcode`. Added an additional feature `extractcode-full` that adds 7zip and libarchive binaries as dependencies instead of needing to rely on them existing on the PATH. Ideally, we'd want to have this hook run *last*, but I don't think pluggy supports that. I also injected .wim support to extractcode just to see what Surfactant can do a Windows 11 iso.

### Future Work

Seeing how extractcode identifies binaries, it seems to use [typecode](https://pypi.org/project/typecode), which internally uses the same library that the linux `file` command uses. This may be a better approach to ID-ing files than what we're currently doing with manually checking header bytes (maybe only do this with more specific or lesser-known file types?). Depending on how extractcode behaves, we may want to also look at removing some of the tar/gz/zip/bz file extraction and letting extractcode handle it (unless we have a reason to look at the file metadata directly).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -59,7 +59,6 @@ dependencies = [
     # Pinned to specific version for potential breaking changes
     "networkx>=2.6",
     "python-msi==0.0.0a2",
-    "extractcode>=31.0.0",
 ]
 
 dynamic = ["version"]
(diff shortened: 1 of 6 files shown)


nightlark (commented):
I think this PR mostly needs conflicts resolved from the file type id changing to support a list of potential types.

From what I recall, we'd also talked about making a change to the upstream extractcode project?

PR #451, opened 2025-07-14
TUI Plugin Settings

KendallHarterAtWork:
### Summary

If merged this pull request will add an option to the TUI that allows changing settings of plugins.

### Proposed changes

This change requires plugins do to two things to have the setting be retrieved automatically:
1. Implement the `settings_name` hook function, returning the section of the config to retrieve settings from.
2. Add file-level doc-strings with a `Config Options:` section


nightlark:
Two changes:
* Extract default value [default=...] from docstrings
* Wrap help text to fit width

surfactant/cmd/tui.py
```diff
@@ -437,6 +454,91 @@ def load(self):
...
+    # Core settings - I don't think there's a good way to automatically extract these
+    __setting = surfactant.utils.get_plugin_settings.PluginSetting
+    __core_settings = [
+        __setting(
+            "output_format",
+            "str",
+            "SBOM output format, see --list-output-formats for list of options",
+            "CyTRICS",
+        ),
+        __setting("recorded_institution", "str", "Name of user's institution.", "LLNL"),
```
nightlark, line 508:
```suggestion
        __setting("recorded_institution", "str", "Name of user's institution.", ""),
```

Default recorded institution as empty string, since theoretically all external users will want something else.

surfactant/cmd/tui.py
```diff
@@ -437,6 +454,91 @@ def load(self):
...
+
+class PluginSettingsTab(textual.widgets.Static):
+    __config_manager = surfactant.configmanager.ConfigManager()
+    # Core settings - I don't think there's a good way to automatically extract these
+    __setting = surfactant.utils.get_plugin_settings.PluginSetting
+    __core_settings = [
+        __setting(
+            "output_format",
+            "str",
+            "SBOM output format, see --list-output-formats for list of options",
```
nightlark, line 505:
```suggestion
            "SBOM output format, see 'surfactant generate --list_output_formats' for list of options",
```

surfactant/cmd/tui.py
```diff
@@ -452,17 +554,19 @@ def __init__(self):
...
-        Tab = textual.widgets.Tab
         with textual.widgets.TabbedContent():
             with textual.widgets.TabPane("Generate"):
                 yield self.generate_tab
             with textual.widgets.TabPane("Merge"):
                 yield self.merge_tab
+            with textual.widgets.TabPane("Plugin Settings"):
+                yield self.plugin_settings_tab
             with textual.widgets.TabPane("Context"):
                 yield self.context_tab
```
nightlark, line 571:
```suggestion
            with textual.widgets.TabPane("Context"):
                yield self.context_tab
            with textual.widgets.TabPane("Settings"):
                yield self.plugin_settings_tab
```

Renaming the tab to just Settings, since it also includes "core" settings. Slight reordering of the tabs to place it at the end.

nightlark (approved):

PR #456, opened 2025-07-15
RPM Package Metadata Extraction

matthewkelley22:
### Summary

If merged this pull request will build on PR #443 by adding metadata extraction for RPM packages.

### Proposed changes

- Adds new dependency on the [rpmfile module](https://github.com/srossross/rpmfile). 
- Adds a new file in the infoextractor directory for rpm files.
- The changes to id_magic.py are the same as  #443.
- Modifies manager.py to include the new rpm_file info extractor.


surfactant/infoextractors/rpm_file.py
```diff
@@ -0,0 +1,147 @@
...
+def get_files(
+    directories: List[bytes], files: List[bytes], indicies: List[int], hashes: List[bytes]
+) -> Dict[Any, Any]:
+    """
+    Extracts files from the given directories and files list
+
+    :param directories: List of directory paths in bytes
+    :param files: List of file names in bytes
+    :param indicies: Directory associated with current file
+    :return: List of extracted file paths
```
nightlark, line 33: For consistency, let's use google style for the docstrings.

surfactant/infoextractors/rpm_file.py
```diff
@@ -0,0 +1,150 @@
...
+    software_field_hints: List[Tuple[str, object, int]],
+    # context_queue: "Queue[ContextEntry]",     # Present for use in future extraction implementation
+    # current_context: Optional[ContextEntry],
+) -> object:
+    if not supports_file(filetype):
+        return None
+    rpm_info = extract_rpm_info(filename)
+
+    for key in rpm_info["rpm"]:
+        software_field_hints.append((key, rpm_info["rpm"][key], 80))
```
nightlark, line 91: I think this works okay because the names we are trying to provide as hints happen to have the same names as checked in https://github.com/LLNL/Surfactant/blob/main/surfactant/cmd/generate.py#L105-L12 checks. Giving the rest of the fields as hints might be a bit pointless, since they're just ignored.

PR #457, opened 2025-07-16
Cache extractions of unsuccessful runs; allow extraction persistence

WorkingRobot:
### Summary

If merged this pull request will allow caching extractions of unsuccessful runs (by default) and allow extractions to be persisted across runs (opt-in).

### Proposed changes

The config file for storing persisted directories is in `%TEMP%/.surfactant_extracted_dirs.json` (or the relevant directory in other platforms. Being located in the same folder as the temp directories, it should allow the end user to clear their temp directory without issue. The config option `decompression.cache_extractions` (True by default) enables the behavior in this PR, while `decompression.persist_extractions` (False by default) enables the behavior even when Surfactant exits successfully. This PR is especially nice when dealing with large archives (i.e. a 5 GB .wim file that Surfactant will be able to handle in #450 that takes 2 hours to extract).


surfactant/infoextractors/file_decompression.py
```diff
@@ -87,19 +100,32 @@ def create_extraction(
...
+    if (
+        software.sha256 in EXTRACT_DIRS
+        and EXTRACT_DIRS[software.sha256]["result"]
+        and os.path.exists(EXTRACT_DIRS[software.sha256]["path"])
+    ):
+        entries = EXTRACT_DIRS[software.sha256]["result"]
+        logger.info(f"Using cached extraction entries for {filename}")
+    else:
+        # Create a temporary directory for extraction
+        temp_folder = create_temp_dir()
```
nightlark, line 112: Can we also make a configuration option for the user to set a specific directory to use instead of a temp dir?

It would also probably be good to add a small addition to the docs page for the available settings to list these new options related to extraction/decompression.

nightlark (approved):

PR #459, opened 2025-07-17
feat(sbom): symlink-aware SBOM filesystem graph (fs_tree)

### LINKED ISSUE #399: Represent filesystem structure as a graph
When gathering information on files to generate an SBOM, store information on the folders and subfolders/files/symlinks as a graph.

### LINKED ISSUE #401: Update relationship establishing plugins to use filesystem graph directly
Update the various relationship establishing plugins to walk the filesystem structure graph directly instead of parsing install/container paths.

willis89pr:
# Symlink-aware SBOM filesystem graph; fs_tree lookup across relationship plugins; path utils & tests

## Summary

This PR makes relationship resolution symlink-aware and more accurate by introducing a first-class filesystem graph (`fs_tree`) inside the SBOM model and teaching the .NET/ELF/PE/Java plugins to resolve dependencies via exact path lookups before falling back to legacy heuristics. It also adds small ergonomics (path utils), targeted logging, safer error handling, performance improvements, and a comprehensive test suite.

## Motivation

Encoding the install tree and symlink edges in a graph lets us: (1) resolve by canonical path, (2) follow links deterministically, and (3) avoid spurious edges.

### 1) SBOM model: new `fs_tree` and helper APIs

* Add `fs_tree: nx.DiGraph` tracking directory hierarchy and **symlink edges** (`type="symlink"`, optional `subtype="file|directory"`).
* New path and lookup helpers:

  * `_add_software_to_fs_tree()` builds path hierarchy and tags nodes with `software_uuid`.
  * `get_software_by_path()` normalizes paths and resolves entries via `fs_tree` with symlink traversal.
  * `get_symlink_sources_for_path()` performs reverse traversal to find all symlinks pointing to a given target.
  * `record_symlink()`, `_add_symlink_edge()`, and `expand_pending_dir_symlinks()` / `expand_pending_file_symlinks()` handle immediate and deferred symlink creation.
  * `record_hash_node()` and `get_hash_equivalents()` track content-equivalent files via SHA-256 nodes.
  * `inject_symlink_metadata()` regenerates legacy-style `fileNameSymlinks` and `installPathSymlinks` fields from the graph (based on symlink edges and hash equivalents only - removed unreachable filename alias logic).
  * `_rebuild_fs_tree_from_metadata()` reconstructs symlink edges from persisted metadata after JSON deserialization, ensuring fs_tree lookups work correctly in merge workflows.
* Extend `add_software_entries()` to merge duplicates, attach `Contains` edges, and link identical hashes.
  * **Removed filesystem probing** - symlink discovery now happens only during extraction/crawl in `generate.py`, not during SBOM operations.
* Split graph builders: `build_rel_graph()` for logical relationships; `fs_tree` is kept separate; **filter out** `path`/`symlink` edges from `to_dict_override()`.
* **Performance**: Optimized `to_dict_override()` to avoid deep-copying large NetworkX graphs before exclusion.
* Standardized graph edge/node type strings to lowercase (`"path"`, `"hash"`, `"symlink"`).
* Added docstrings and safety checks across all new helpers.

### 2) `generate.py`: symlink capture during crawl

* Inject filename- and install-path symlinks into `Software` entries before adding them.
* Record hash-based equivalence edges for symlink pairs.
* After crawl, expand deferred symlinks and inject legacy-style symlink metadata using `inject_symlink_metadata()`.
* Debug logs added; broad exceptions tightened; duplicate logging removed.

### 3) Relationship plugins

* **.NET** (`dotnet_relationship.py`):

  * Unified path normalization with `normalize_path`.
  * Resolve via `sbom.get_software_by_path`, handle `app.config` (`probing.privatePath`, `<codeBase href=...>`), unmanaged imports via `dotnetImplMap`.
  * Added **culture/version filtering**; clarified that Culture is used for filtering, not probing subdirs.
  * Fixed probe directory construction to normalize consistently.
  * **Added guard**: `has_required_fields()` now checks `isinstance(metadata, dict)` to prevent TypeError.

* **ELF** (`elf_relationship.py`):

  * Introduced structured docstrings and debug logs.
  * RPATH/RUNPATH resolution respects `DF_1_NODEFLIB`.
  * Expanded DST substitution with `$ORIGIN` and `$LIB`.
  * Phased resolution: fs_tree → legacy → heuristic.
  * Standardized final-emission debug logs and clarified when no match is found.
  * Condensed `DF_1_NODEFLIB` check and explicitly logs default search paths.
  * Replaced ad-hoc `print` with `logger.debug` for expanded runpaths.
  * **Added guard**: `has_required_fields()` now checks `isinstance(metadata, dict)`.

* **PE** (`pe_relationship.py`):

  * Uses fs_tree → legacy → same-directory match (Phase 2/3).
  * Clarified test case docstring and renamed misleading "symlink heuristic" test to **same-directory match**.
  * **Added guard**: `has_required_fields()` now checks `isinstance(metadata, dict)`.

* **Java** (`java_relationship.py`):

  * Corrected Phase 1 test to align importer path with fs_tree lookups.
  * Corrected Phase 3 heuristic test to require same parent dir + filename, with Phase 2 failing.
  * **Note**: fs_tree lookup not yet implemented for Java (legacy matching only).

### 4) Merge and graph hygiene

* Skip `path` nodes in root computation, merges, and relationship output.

### 5) New path utilities

* `surfactant/utils/paths.py`

  * `normalize_path(*parts) → str` ensures consistent POSIX normalization across Windows/Unix.
  * `basename_posix(path) → str`.

### 6) Tests

* **.NET**:

  * Removed redundant manual `fs_tree.add_node()`.
  * Clarified `test_dotnet_culture_subdir` docstring (filtering only).
  * Added `test_dotnet_heuristic_match` for Phase 3.

* **ELF**:

  * Switched from private `_record_symlink` to public API; expanded docstrings for clarity.
  * Updated example fixture with explicit consumer to exercise system-path fallback.
  * Added `test_symlink_heuristic_match_edge` to force heuristic after clearing direct symlink edge.

* **PE**:

  * Cleaned up commented fs_tree code.
  * Renamed and clarified same-directory test; added explicit path normalization assert.

* **Java**:

  * Corrected Phase 1 test to actually hit fs_tree resolution.
  * Corrected Phase 3 test to properly require same-dir heuristic.

* Added `tests/utils/test_paths.py` for `normalize_path` (including test name fix: `test_trailing_slash_is_stripped`).

* Added `tests/sbomtypes/test_fs_tree.py` to validate `fs_tree` population, lookup, and **reconstruction from metadata after deserialization**.

### 7) Bug fixes and improvements

* Fixed missing `pathlib` import in `_sbom.py`.
* Removed duplicate logging in `expand_pending_dir_symlinks()`.
* Standardized node/edge type strings from uppercase to lowercase.
* Removed unreachable `file_name_extras` logic from `inject_symlink_metadata()`.
* Fixed filesystem probing in `add_software_entries()` that caused nondeterministic behavior during merges.

## Risk & compatibility

* **Behavioral improvements only**: JSON outputs exclude path/symlink edges.
* `fs_tree` is internal (non-serialized).
* Symlinks are persisted in metadata and reconstructed on load.

## Performance considerations

* Linear path-tree construction, O(1) symlink recording.
* Faster lookups due to direct node queries; legacy/heuristic only if needed.
* **Optimized serialization**: `to_dict_override()` no longer deep-copies large graphs.

## Logging & DX

* Concise `logger.debug` traces for resolution phases.
* Failures to record symlinks warn, not error.
* Duplicate logging removed.

## SBOM model note (post-deserialization)

* After loading from JSON, `SBOM.__post_init__()` calls `_rebuild_fs_tree_from_metadata()` to reconstruct symlink edges from `installPathSymlinks` metadata, ensuring `fs_tree` lookups work correctly across load/merge workflows.
* **No filesystem probing** occurs during SBOM operations - all symlink information comes from metadata captured during extraction.

## Test plan

```bash
pytest -q tests/relationships/test_dotnet_relationship.py
pytest -q tests/relationships/test_elf_relationship.py
pytest -q tests/relationships/test_java_relationship.py
pytest -q tests/relationships/test_pe_relationship.py
pytest -q tests/utils/test_paths.py
pytest -q tests/sbomtypes/test_fs_tree.py
pytest -q  # full suite (137 passing, 4 skipped)
```


nightlark:
9/22 - working on tracking down a bug that results in not producing an SBOM

surfactant/relationships/java_relationship.py
nightlark: I don't think the java relationships were established using paths at all previously. Can we split the java changes out into a separate branch to punt it to later? I need to dig a bit more into how Java finds classes.

surfactant/relationships/pe_relationship.py
```diff
@@ -4,75 +4,209 @@
...
+            for item in sbom.software:
+                # Need a name match first
+                has_name = isinstance(item.fileName, Iterable) and fname in (item.fileName or [])
+                if not has_name:
+                    continue
+
+                # Then ensure any of the item's install paths share a probedir dir
+                if isinstance(item.installPath, Iterable):
+                    for ipath in item.installPath or []:
+                        ip_dir = pathlib.PurePosixPath(ipath).parent.as_posix()
```
nightlark: This should probably be a PureWindowsPath instead for a case-insensitive comparison.

surfactant/relationships/pe_relationship.py
```diff
@@ -4,75 +4,209 @@
...
+        # ---------------------------------------------------------
+        if not matched_uuids:
+            for item in sbom.software:
+                has_name = isinstance(item.fileName, Iterable) and fname in (item.fileName or [])
+                if not has_name or not isinstance(item.installPath, Iterable):
+                    continue
+                for ipath in item.installPath or []:
+                    ip_dir = pathlib.PurePosixPath(ipath).parent
+                    for refdir in probedirs:
+                        if ip_dir == pathlib.PurePosixPath(refdir):
```
nightlark: I'm not seeing how this check is different than Phase 2.

It looks like the file name check beforehand is essentially the same, and this looks like roughly the equivalent of `if ip_dir in probedirs:` but with an extra loop.

surfactant/relationships/elf_relationship.py
```diff
@@ -42,107 +44,173 @@ def has_required_fields(metadata) -> bool:
...
+                for ipath in item.installPath or []:
+                    ip_dir = pathlib.PurePosixPath(ipath).parent
+                    for fp in fpaths:
+                        if pathlib.PurePosixPath(fp).parent == ip_dir:
+                            if item.UUID != software.UUID:
+                                logger.debug(
+                                    f"[ELF][heuristic] {fname} via {ipath} ~ {fp} → UUID={item.UUID}"
+                                )
+                                matched_uuids.add(item.UUID)
+                                used_method[item.UUID] = "heuristic"
```
nightlark: Phase 3 is essentially doing the same thing as Phase 2 except without verifying that the file name at the end of the install path is correct, opening up the possibility of false positives. For a dependency of "somedir/abc", the full path must match, which includes the file name. With this heuristic, we end up with a scenario where a software entry like this would create a false positive:

file names = ["abc", "def"]
install paths = ["somedir/def", "anotherdir/abc"]

No match should be found because "abc" doesn't exist in "somedir" (perhaps an indication that something is broken on the system we got the filesystem dump from, but we aren't trying to fix a broken device).

surfactant/relationships/elf_relationship.py
```diff
@@ -42,107 +44,173 @@ def has_required_fields(metadata) -> bool:
...
 
 def generate_search_paths(sw: Software, md) -> List[pathlib.PurePosixPath]:
-    # 1. Search using directories in DT_RPATH if present and no DT_RUNPATH exists (use of DT_RPATH is deprecated)
-    # 2. Use LD_LIBRARY_PATH environment variable; ignore if suid/sgid binary (nothing to do, we don't have this information w/o running on a live system)
-    # 3. Search using directories in DT_RUNPATH if present
-    paths = generate_runpaths(sw, md)  # will return an empty list if none
-
-    # 4. From /etc/ld.so.cache (/var/run/ld.so.hints on FreeBSD) list of compiled candidate libraries previously found in augmented library path; if binary was linked with -z nodeflib linker option, libraries in default library paths are skipped
-    # /etc/ld.so.conf can be used to add additional directories to defaults (e.g. /usr/local/lib or /opt/lib), but we don't necessarily have a way to gather this info
-    # Search in default path /lib, then /usr/lib; skip if binary was linked with -z nodeflib option
```
nightlark: I think keeping the numbered steps comments in roughly the same places would be useful (they are roughly pulled from docs in some of the Linux man pages).

surfactant/relationships/elf_relationship.py
```diff
@@ -173,50 +241,113 @@ def replace_dst(origstr, dvar, newval) -> str:
     return origstr.replace("$" + dvar, newval).replace("${" + dvar + "}", newval)
 
 
+# def substitute_all_dst(sw: Software, md, path) -> List[pathlib.PurePosixPath]:
```
nightlark: Instead of keeping the entire commented out function, let's just keep the comments in it as a way to capture the summary of info on DSTs and where different bits of info come from.

surfactant/relationships/elf_relationship.py
willis89pr: Okay, I brought the comments back in.

surfactant/relationships/dotnet_relationship.py
```diff
@@ -1,243 +1,229 @@
...
+
+                for sw_path in sw.installPath:
+                    sw_dir = pathlib.PurePath(sw_path).parent
+                    for refdir in probedirs:
+                        if sw_dir == pathlib.PurePath(refdir):
+                            logger.debug(
+                                f"[.NET][heuristic] {refName} via {sw_path} → UUID={sw.UUID}"
+                            )
+                            matched_uuids.add(sw.UUID)
+                            used_method[sw.UUID] = "heuristic"
```
nightlark: I think similar to some of the other instances, the shared directory heuristic here will in some edge cases add potential for false positives that didn't exist before, while the non-false positive cases look like they'd still be covered by phase 2.

surfactant/sbomtypes/_sbom.py
```diff
@@ -216,51 +424,303 @@ def add_software(self, sw: Software) -> None:
...
+        """
+        Add software entries to the SBOM graph, merging duplicates, preserving existing edges,
+        attaching "Contains" relationships to an optional parent, and recording ANY
+        file- or directory-level symlinks under each installPath.
 
         Args:
-            entries (Optional[List[Software]]): A list of Software entries to add to the SBOM.
-            parent_entry (Optional[Software]): An optional parent software entry to add "Contains" relationships to.
+            entries:        list of Software instances to add.
+            parent_entry:   if provided, attach a "Contains" edge from this parent to each entry.
```
nightlark: Is there a reason type information is being stripped from the docstrings?

surfactant/relationships/elf_relationship.py
```diff
@@ -42,107 +44,173 @@ def has_required_fields(metadata) -> bool:
...
     relationships: List[Relationship] = []
     dependent_uuid = software.UUID
     default_search_paths = generate_search_paths(software, metadata)
+    logger.debug(f"[ELF][search] default paths: {[p.as_posix() for p in default_search_paths]}")
+
     for dep in metadata["elfDependencies"]:
-        # if dependency has a slash, it is interpreted as a pathname to shared object to load
-        # construct fname and full file path(s) to search for; paths must be a list if the dependency is given as a relative path
+        fpaths = []
+        dep_path = posix_normpath(dep)
```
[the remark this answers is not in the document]
willis89pr: The normalization was already happening inside the `if "/" in dep` branch. Hoisting `posix_normpath(dep)` to `dep_path` and using it in both branches doesn’t alter the slash-based case selection. 

Here is a more detailed breakdown:
### 1. What does `posix_normpath("ab/../xy")` do?

Given your implementation:

```python
posix_path = pathlib.PurePosixPath(path)
parts = list(posix_path.parts)
...
if parts[i] == "..":
    del parts[i]
    if i > 0:
        if i > 1 or parts[0] not in ("//", "/"):
            del parts[i - 1]
            i -= 1
...
return pathlib.PurePosixPath(*parts)
```

For `"ab/../xy"`:

* `posix_path.parts` → `("ab", "..", "xy")`
* At `i = 1`, `parts[1] == ".."`, so you delete `".."` and then delete the preceding `"ab"`.
* Resulting `parts` → `["xy"]`
* Return `PurePosixPath("xy")`.

So for `"ab/../xy"`:

```python
dep_path = posix_normpath("ab/../xy")  # → PurePosixPath("xy")
dep_path.name                          # → "xy"
dep_path.is_absolute()                 # → False
dep_path.as_posix()                    # → "xy"
```

Exactly what we thought.

### 2. Old vs new: which case do we take?

**New code:**

```python
for dep in metadata["elfDependencies"]:
    fpaths = []
    dep_path = posix_normpath(dep)
    fname = dep_path.name  # 'xy'

    # Case 1: Dependency has slash — treat as direct path
    if "/" in dep:
        if dep_path.is_absolute():
            fpaths = [dep_path.as_posix()]
        else:
            ...
```

Condition for Case 1 is **still**:

```python
if "/" in dep:
```

That checks the original string, not the normalized `dep_path`.

For `dep = "ab/../xy"`:

* `"/" in dep` → **True**
* We go into Case 1.
* `dep_path.is_absolute()` is False, so we take the **relative-path sub-branch**, same as legacy.

**Legacy code:**

```python
for dep in metadata["elfDependencies"]:
    if "/" in dep:
        dep = posix_normpath(dep)
        fname = dep.name  # 'xy'
        if dep.is_absolute():
            fpaths = [str(dep)]
        else:
            fpaths = []
            if isinstance(software.installPath, Iterable):
                for ipath in software.installPath:
                    ipath_posix = posix_normpath(ipath)
                    fpaths.append(
                        posix_normpath(str(ipath_posix.parent.joinpath(dep))).as_posix()
                    )
```

Again:

* `"/" in "ab/../xy"` → **True** → Case 1.
* `dep = posix_normpath(dep)` → `PurePosixPath("xy")`
* `dep.is_absolute()` → False → relative-path sub-branch.

So **in both versions**:

* `"ab/../xy"` is treated as a “has slash ⇒ direct path” dependency.
* It never falls through to the “bare filename” / Case 2 behavior.

That directly contradicts the reviewer’s worry (“normpath will return xy, so instead case 2 will get used”).

### 3. Any subtle behavior changes?

Let’s compare the relative-path `fpaths` construction, since that’s the only real potential difference.

**Legacy:**

```python
ipath_posix = posix_normpath(ipath)
fpaths.append(
    posix_normpath(str(ipath_posix.parent.joinpath(dep))).as_posix()
)
```

Here `dep` is already normalized (`PurePosixPath("xy")`), so:

* `ipath_posix = posix_normpath(ipath)`
* `ipath_posix.parent.joinpath(dep)` → `<parent_of_ipath>/xy`
* `posix_normpath(str(...))` normalizes again.

**New:**

```python
ipath_posix = posix_normpath(ipath)
combined = ipath_posix.parent.joinpath(dep_path)
fpaths.append(combined.as_posix())
```

Differences:

* You no longer do a second `posix_normpath` on the combined path. But since:

  * `ipath_posix = posix_normpath(ipath)` (already normalized),
  * `dep_path = posix_normpath(dep)` (already normalized, no `..` left),
    the combined path doesn’t contain extra `".."` components to collapse. The second normalization was effectively a no-op for normal inputs.

So:

* Case selection: **unchanged**.
* `fname` for `"ab/../xy"`: **unchanged** (`"xy"`).
* Absolute vs relative handling: **unchanged**.
* Generated `fpaths` values for typical inputs: **effectively unchanged**.

### 4. Direct answer to your concern

* The specific scenario (“`ab/../xy` will now be treated as a bare filename and hit Case 2”) is **incorrect**.
* Your refactor does **not** change whether `"ab/../xy"` goes through Case 1 vs Case 2. It still goes through Case 1 (direct path) in both old and new code.

surfactant/relationships/pe_relationship.py
willis89pr: I removed Phase 3 in pe_relationship.py

surfactant/relationships/pe_relationship.py
```diff
@@ -4,75 +4,209 @@
...
-            relationships.append(Relationship(dependent_uuid, dependency_uuid, "Uses"))
-        # logging DLLs not found would be nice, but is excessively noisy due being almost exclusively system DLLs
+            for ipath in sw.installPath or []:
+                parent_dir = pathlib.PureWindowsPath(ipath).parent.as_posix()
+                probedirs.append(parent_dir)
+        logger.debug(f"[PE][import] probedirs for '{fname}': {probedirs}")
+
+        for directory in probedirs:
+            full_path = normalize_path(directory, fname)
+            match = sbom.get_software_by_path(full_path)
```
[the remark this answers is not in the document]
willis89pr: Currently, `get_software_by_path` is case-sensitive. `normalize_path` only normalizes path separators, and the `fs_tree` lookup is an exact string match on the normalized path, so Windows paths with different casing won’t match.

surfactant/relationships/pe_relationship.py
willis89pr: I switched the legacy path handling from `PurePosixPath` to `PureWindowsPath` when computing `ip_dir`

surfactant/relationships/pe_relationship.py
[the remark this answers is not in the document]
willis89pr: I added a case-insensitive check for windows in `sbom.get_software_by_path()`

surfactant/relationships/java_relationship.py
willis89pr: I made sure Phase 2 has all of the legacy functionality, then commented out Phase 1.

surfactant/relationships/dotnet_relationship.py
```diff
@@ -1,243 +1,586 @@
...
+# Feature flag: enable full legacy probe (O(n^2) fallback) when explicitly requested.
+# This is used later as the "no compromises"
+# TODO: Make this a CLI flag or config option instead of env var
+# FULL_DOTNET_LEGACY_PROBE = os.getenv("SURFACTANT_DOTNET_FULL_PROBE", "").lower() in (
+#     "1",
+#     "true",
+#     "yes",
+#     "on",
+# )
+FULL_DOTNET_LEGACY_PROBE = 0
```
nightlark: can remove feature flag

tests/symlink/test_resolve_links.py
nightlark: Why is this file getting removed? The function being tested still exists.

surfactant/sbomtypes/_sbom.py
```diff
@@ -379,6 +1186,14 @@ def merge(self, sbom_m: SBOM):
 
         # 3) Merge relationships from the incoming SBOM’s MultiDiGraph
         for src, dst, rel_type in sbom_m.graph.edges(keys=True):
+            # Skip path/symlink edges during merge as well
+            if str(rel_type).lower() == "symlink":
+                continue
+            if sbom_m.graph.nodes.get(src, {}).get("type") == "Path":
```
nightlark: Should we make all of the node types lower case for consistency ("path" and "hash" instead of "Path" and "Hash")?

surfactant/relationships/elf_relationship.py
```diff
@@ -59,93 +61,200 @@ def establish_relationships(
...
     Notes:
-        - The function uses `metadata["elfDependencies"]` to locate dependencies described
-          as ELF paths or filenames.
-        - Relative paths in metadata are normalized and matched against installation paths
-          of the candidate software entries.
-        - Dependency file paths are cross-referenced with `sbom.software` entries to establish
-          their relationships.
-        - Returned `Relationship` objects are unique: no duplicates are added to the result list.
+        - Dependency paths may originate from:
+            • absolute paths in metadata
```
nightlark: Might be better to keep the docstrings to plain ascii (or markdown-ish) to avoid potential weird stuff happening if we add automated doc generation for the website.

surfactant/relationships/elf_relationship.py
```diff
@@ -180,76 +289,27 @@ def generate_search_paths(sw: Software, md) -> List[pathlib.PurePosixPath]:
...
-        ]
-        ```
-        And `sw` enables substitution tokens such as `$LIB`.
-        The function will return resolved paths by splitting `"/lib:/usr/lib"`
-        and applying substitutions where `$LIB` is located.
-
-    Notes:
-        - If the ELF file specifies both `DT_RPATH` and `DT_RUNPATH`,
-          `DT_RUNPATH` is given precedence.
-    """
```
nightlark: Probably should keep this comment

surfactant/sbomtypes/_sbom.py
```diff
@@ -216,51 +489,585 @@ def add_software(self, sw: Software) -> None:
...
+            1. **Filesystem Symlinks:** incoming symlink edges in `fs_tree`
+               (e.g., "usr/sbin/runuser" → "usr/bin/su")
+
+            2. **Hash-Equivalent Siblings:** other files that share identical content
+               (sha256) but appear at different install paths.
+
+            3. **Gathered Filename Aliases:** additional names from `sw.fileName`
+               that were injected during the gather phase but are not canonical
+               basenames of the install paths (e.g., bash-completion stubs like
+               "runuser" for "su").
```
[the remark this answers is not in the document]
willis89pr: resolved in (commit 3c4d863)

nightlark (approved):

PR #475, opened 2025-08-11
refactor(test_grype): simplify install_grype() to use Anchore endpoint

willis89pr:
**Description:**
This pull request streamlines our Grype installation helper in `test_grype.py` by switching from the brittle GitHub‐raw installer to Anchore’s official “get.anchore.io/grype” endpoint. In doing so, we:

* Remove the manual download of `install.sh`, the `chmod` step, and cleanup logic
* Eliminate the unused `version` parameter and related branching
* Pipe the installer directly into `sudo sh` for a one-step, reliable install of the latest release
* Keep the pre-install check (`which("grype")` + `grype --version`) to skip if already present
* Retain post-install verification via `grype --version` and appropriate logging

**Why:**

* **Reliability:** The old two-step script frequently failed in CI due to GitHub’s “latest” redirect returning 503, leaving no binary on `$PATH`.
* **Simplicity:** Piping `curl | sh` removes intermediate files and filesystem noise (no temporary `install.sh` to chmod or delete).
* **Maintainability:** Always installing the latest stable Grype avoids stale version drift, and uses an officially supported endpoint.

**Testing:**

1. Run `install_grype()` locally on a system without Grype to confirm it fetches and installs the binary.
2. Verify that `grype --version` succeeds both before (if previously installed) and after the helper runs.
3. Trigger the CI pipeline to ensure `test_surfactant_generate` no longer errors out with `FileNotFoundError`.

---

### DIFF
--- a/plugins/grype/tests/test_grype.py
+++ b/plugins/grype/tests/test_grype.py
@@ -67,10 +67,7 @@ def install_grype() -> None:
 
     logging.info("Installing Grype via get.anchore.io…")
     # Use the official Anchore installer to fetch and install the latest Grype
-    cmd = (
-        "curl -sSfL https://get.anchore.io/grype"
-        " | sudo sh -s -- -b /usr/local/bin"
-    )
+    cmd = "curl -sSfL https://get.anchore.io/grype | sudo sh -s -- -b /usr/local/bin"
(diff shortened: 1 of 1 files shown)


willis89pr:
Grype updated there install instructions: https://github.com/anchore/grype/pull/2833

PR #478, opened 2025-08-14
SBOM Visualization Plugin

JosephHAtWork:
### Summary

If merged this pull request will add a plugin that provides interactive visualizations for Surfactant SBOMs

### DIFF
--- a/plugins/SBOMVis/pyproject.toml
+++ b/plugins/SBOMVis/pyproject.toml
@@ -3,13 +3,13 @@ requires = ["setuptools", "setuptools-scm"]
 build-backend = "setuptools.build_meta"
 
 [project]
-name = "surfactant-SBOMVis"
+name = "sbomvis"
 authors = [
     {name = "Joseph Hand", email = "hand8@llnl.gov"},
 ]
 description = "Surfactant SBOM Visualization"
 readme = "README.md"
-requires-python = ">=3.8"
+requires-python = ">=3.9"
 keywords = ["surfactant"]
 license = {text = "MIT License"}
 classifiers = [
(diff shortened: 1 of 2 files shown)


plugins/SBOMVis/pyproject.toml
```diff
@@ -0,0 +1,40 @@
+[build-system]
+requires = ["setuptools", "setuptools-scm"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "surfactant-SBOMVis"
```
nightlark, line 6:
```suggestion
name = "sbomvis"
```

Let's make the name sbomvis -- once Surfactant gets the ability to read CycloneDX/SPDX SBOM files as input then the visualizer will be pretty close to usable with SBOMs generated by other tools.

Description can be updated later.

plugins/SBOMVis/pyproject.toml
```diff
@@ -0,0 +1,40 @@
...
+authors = [
+    {name = "Joseph Hand", email = "hand8@llnl.gov"},
+]
+description = "Surfactant SBOM Visualization"
+readme = "README.md"
+requires-python = ">=3.8"
```
nightlark, line 12:
```suggestion
requires-python = ">=3.9"
```

Surfactant requires Python >=3.9, so that's a transitive dependency on a newer version of Python for sbomvis.

nightlark (commented):
I added the example image and a smaller version to https://github.com/LLNL/Surfactant/issues/488 so we can use links to the images hosted on GitHub instead of adding a bit of a larger "binary" file to the git repo.

PR #499, opened 2025-09-16
Add DAPper plugin

monwen:
### Summary

### DIFF
--- a/plugins/dapper/README.md
+++ b/plugins/dapper/README.md
@@ -19,29 +19,22 @@ The Dapper plugin enhances Surfactant's SBOM generation by identifying which sof
 
 Before using this plugin, you need to install Dapper and download the package databases:
 
-1. Install dapper-python (required for database queries):
+1. Install Dapper:
    ```bash
-   pip install dapper-python
+   cargo install dapper
    ```
 
-2. Clone and build Dapper:
-   ```bash
-   git clone https://github.com/LLNL/dapper.git
-   cd dapper
-   cargo build --release
-   ```
-
-3. Download required datasets (must be run from the dapper directory):
+2. Download required datasets (must be run from the dapper directory):
    ```bash
    # List available datasets
    cargo run -- . --list-available-datasets
-   
+
    # Install specific Linux datasets
    cargo run -- . --install ubuntu-focal
    cargo run -- . --install ubuntu-jammy
    cargo run -- . --install ubuntu-noble
    cargo run -- . --install debian-bookworm
-   
+
    # Install all available datasets
    cargo run -- . --install all
    ```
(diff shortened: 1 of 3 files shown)


plugins/dapper/README.md
```diff
@@ -0,0 +1,168 @@
...
+## Prerequisites
+
+### Install Dapper and Datasets
+
+Before using this plugin, you need to install Dapper and download the package databases:
+
+1. Install dapper-python (required for database queries):
+   ```bash
+   pip install dapper-python
+   ```
```
nightlark, line 25: If dependencies were listed properly in pyproject.toml for the plugin, this step should be removed.

plugins/dapper/README.md
```diff
@@ -0,0 +1,168 @@
...
+
+   # Install specific Linux datasets
+   cargo run -- . --install ubuntu-focal
+   cargo run -- . --install ubuntu-jammy
+   cargo run -- . --install ubuntu-noble
+   cargo run -- . --install debian-bookworm
+
+   # Install all available datasets
+   cargo run -- . --install all
+   ```
```
nightlark, line 47: This can be updated later to use `cargo install dapper`, so no cloning is required, and then the user will have a binary on their path that lets them use `dapper` instead of `cargo run --` in all the subsequent commands for installing datasets.

plugins/dapper/README.md
```diff
@@ -0,0 +1,168 @@
...
+## Output Format
+
+The plugin adds package information to the metadata field of software entries in the SBOM. For each binary file, it provides:
+
+```json
+{
+  "dapper_packages": [
+    {
+      "package_name": "coreutils",
+      "full_package_name": "coreutils/focal 8.30-3ubuntu2",
```
nightlark, line 76: I think we should remove the full_package_name, the version information is misleading (we don't have any indication about what the version number is or should be).

plugins/dapper/README.md
```diff
@@ -0,0 +1,168 @@
...
+```bash
+pipx uninject surfactant surfactantplugin-dapper
+```
+
+## License
+
+MIT License (same as Surfactant)
+
+## Additional Resources
+
```
nightlark, line 164: https://crates.io/crates/dapper can be added to the list of resources for dapper.

plugins/dapper/surfactantplugin_dapper.py
```diff
@@ -0,0 +1,183 @@
...
+
+try:
+    from dapper_python.databases.linuxDB import LinuxDB
+    from dapper_python.dataset_loader import DatasetCatalog
+    from dapper_python.normalize import NormalizedFileName, normalize_file_name
+
+    DAPPER_AVAILABLE = True
+except ImportError:
+    DAPPER_AVAILABLE = False
+    logging.warning("dapper-python not installed. Dapper plugin will be disabled.")
```
nightlark, line 17: If the plugin was pip installed, this dependency should always be available because it is listed as a dependency in pyproject.toml and therefore pip will install it. The only time I can think of it not being present is if someone did some strange manual installation of the plugin by copying it into their site-packages folder and side stepped pip entirely.

PR #502, opened 2025-09-22
Handle RPM md5 fields with other algorithms

matthewkelley22:
### Summary

The md5 fields in RPM packages may be used to hold a sha1 hash, sha256 hash, or a hash with some other algorithm. 

This PR adds support for differentiating between sha1, sha256, sha512, and md5 hashes for:

-  The hash stored in the "md5" field of the RPM header
- Hashes of files stored within the RPM package

### Proposed changes

Modifies the rpm_file infoextractor, and adds a field in the SBOM for:

- What hashing algorithm was used for the files inside the current RPM package

### DIFF
--- a/surfactant/infoextractors/rpm_file.py
+++ b/surfactant/infoextractors/rpm_file.py
@@ -28,9 +28,9 @@ def algo_from_id(algo_identifier: int) -> str:
     Args:
         algo_identifier (int): Integer in RPM file associated with a hashing algorithm
     """
-    if 0 == algo_identifier:
+    if algo_identifier == 0:
         return "md5"
-    if 8 == algo_identifier:
+    if algo_identifier == 8:
         return "sha256"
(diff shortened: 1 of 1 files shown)


surfactant/infoextractors/rpm_file.py
```diff
@@ -22,34 +22,73 @@ def supports_file(filetype) -> bool:
...
+    """Grabs the hashing algorithm used for payload files.
+
+    Args:
+        algo_identifier (int): Integer in RPM file associated with a hashing algorithm
+    """
+    if 0 == algo_identifier:
+        return "md5"
+    if 8 == algo_identifier:
+        return "sha256"
+    return f"Unkown: {algo_identifier}"
```
nightlark, line 36:
```suggestion
    return f"Unknown: {algo_identifier}"

```

PR #507, opened 2025-09-29
Capture ELF dependencies from program headers

heatwole1:
### Summary

If merged this pull request will capture ELF dependencies that only appear in the program headers. This is essential to capture dependencies in ELF files with stripped section headers.

### Proposed changes

1. Change the `elfDependencies` to a set to avoid duplicates if the same reference is present in both SHT_DYNAMIC and PT_DYNAMIC
2. Check the dynamic segment (PT_DYNAMIC) from the ELF program headers in addition to the dynamic section (SHT_DYNAMIC) from the section headers.

Officially the dynamic section and the dynamic segment should be identical, but since they're both essentially just pointers it's possible they can be different. @nightlark suggested scanning both and merging them in a set so that's what this PR does, but I've only tested it on a single sample.

This PR is primarily to draw attention to the issue. I have no strong preference on the implementation.

### DIFF
--- a/surfactant/infoextractors/elf_file.py
+++ b/surfactant/infoextractors/elf_file.py
@@ -63,7 +63,7 @@ def extract_elf_info(filename: str) -> object:
         # Don't assume OS is Linux, map e_ident EI_OSABI value to an OS name
         file_details: Dict[str, Any] = {"OS": ""}
         file_details["elfIdent"] = get_elf_ident_from_file_header(f, elf.little_endian)
-        file_details["elfDependencies"] = []
+        file_details["elfDependencies"] = set()
         file_details["elfRpath"] = []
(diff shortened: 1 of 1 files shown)


nightlark (approved):
Looks good, this should ensure no dependencies are left out.

Since PT_DYNAMIC is what is actually used by the loader, I guess in some ways that is the more important one -- in almost all cases I imagine SHT_DYNAMIC will be equal to or a subset (likely empty set for a stripped binary) of PT_DYNAMIC. A malicious binary might remove a few things from SHT_DYNAMIC to hide stuff from tools, which this change will make sure the actual dependencies still get captured (not sure if its worth warning in this case about the discrepancy); I can't really think of reasons for someone would want to do the opposite and add a fake dependencies to SHT_DYNAMIC.

PR #530, opened 2025-11-04
Update elf_relationship.py

CoWeAtWork:
Adding docstrings to each of the elf_relationship.py functions.

### Summary

If merged this pull request will add docstrings to each of the elf_relationship.py functions. Please check the example sections to make sure that the functions are being represented appropriately.

Additionally, it may be worth merging the comments into the docstrings to take up less space or remove them now that we have the docstrings. However, I wanted to verify the accuracy of the docstrings before altering the comments.

### DIFF
--- a/surfactant/relationships/elf_relationship.py
+++ b/surfactant/relationships/elf_relationship.py
@@ -17,7 +17,7 @@ def has_required_fields(metadata) -> bool:
     Checks if the metadata contains the required `elfDependencies` field.
 
     This function determines whether the `elfDependencies` field exists in the provided
-    metadata. This field indicates the necessary 
+    metadata. This field indicates the necessary
     dependency information to establish relationships.
 
     Args:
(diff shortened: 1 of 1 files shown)


surfactant/relationships/elf_relationship.py
```diff
@@ -13,6 +13,22 @@
...
+    dependency information to establish relationships.
+
+    Args:
+        metadata: The metadata provided to determine the presence of elfDependencies.
+
+    Returns:
+        bool: True if the `elfDependencies` field exists in the metadata, False otherwise.
+
+    Raises:
+        None
```
nightlark: iirc if it doesn't raise any exception then specifying this isn't needed. (actually, it probably could raise an error if the user passes a value for `metadata` that doesn't support using the "in" keyword.

nightlark (approved):

PR #554, opened 2025-12-02
Add Install Prefix option to generate command

JosephHAtWork:
### Summary

If merged this pull request will add an `--install_prefix` option to the `surfactant generate` command. This allows users to quickly generate an SBOM for a filesystem or software installation without needing to create a dedicated configuration file for it. If both a configuration file for the SPECIMEN_CONTEXT and the `--install_prefix` option are provided, the value specified on the command line will override the prefix defined in the configuration file.

### Proposed changes

* Add install prefix option
* Update Basic Usage documentation

### DIFF
--- a/surfactant/cmd/generate.py
+++ b/surfactant/cmd/generate.py
@@ -267,7 +267,7 @@ def get_default_from_config(option: str, fallback: Optional[Any] = None) -> Any:
 )
 @click.option(
     "--install_prefix",
-    "install_prefix_",
+    "install_prefix_arg",
     is_flag=False,
     default=None,
     help="SBOM install prefix",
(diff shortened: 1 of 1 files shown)


surfactant/cmd/generate.py
```diff
@@ -348,13 +357,20 @@ def sbom(
...
                 parent_uuid = None
 
+            # Replace entry install prefix with user specified value if given by cli args
+            if install_prefix_:
+                if entry.installPrefix:
+                    logger.warning(
+                        f"Conflicting installPrefix definitions; Overriding config file ({entry.installPrefix}) with CLI argument ({install_prefix_})"
+                    )
+
+                entry.installPrefix = install_prefix_
```
nightlark: I wonder if it would be safer to say that the `--install_prefix` CLI option can only be used when Surfactant is given a file/folder as input rather than a context file? Then we just can make it an error to give an install prefix flag with the context file and exit before generating an SBOM.

If someone has gone through the effort of creating a context file with install prefixes specified for entries, it seems like overriding them with the CLI option would only ever happen as a mistake.

nightlark (approved):

PR #601, opened 2026-03-24
Align CyTRICS v1.0.1 Schema

willis89pr:
## Summary

This PR updates Surfactant’s CyTRICS data model, serialization shape, fixtures, and tests to align with schema v1.0.1. It adds the v1.0.1 BOM root fields and structured types expected by the schema (`bomUUID`, `bomFormat`, `bomDescription`, `specVersion`, optional `authors`, and optional `tools`), adds support for structured `name`/`comments`, `softwareType`, and relationship comments, switches `captureTime` handling to RFC 3339 strings, and enforces field-level checks through explicit `validate()` methods and schema-focused tests. Runtime construction and serialization remain permissive so schema-invalid input can be carried through and caught by explicit validation or pytest schema validation rather than failing during normal read/write paths.

## What changed

* Updated the typed CyTRICS model and serialization shape for v1.0.1. This adds `Author`, `Tool`, `NameEntry`, and `CommentEntry` types, BOM root fields, `softwareType`, relationship comments, schema-aware omission of selected null optional fields, and RFC 3339 capture-time helpers. Runtime SBOM load/serialization paths now preserve input data instead of performing full schema validation automatically; schema conformance is covered by explicit `validate()` methods and pytest JSON Schema tests.

* Removed legacy CyTRICS constructs from the typed model and merge/generate flow, including `systems`, `analysisData`, `observations`, `starRelationships`, provenance types, `recordedInstitution`, and legacy software component wrappers. The `merge` command and helper scripts no longer support config/system-wrapping behavior, the generate path drops `--recorded_institution`, and the CLI/script paths no longer create or wire in a synthetic top-level system object. The CycloneDX and SPDX writers were updated accordingly so they no longer emit system-derived components/packages from the old model.

* Updated generation and plugin contracts to match the new v1.0.1 data flow. `identify_file_type` can now return either a string or `list[str]`, generator code normalizes file-type results before extraction, extractor hooks are documented to return JSON-compatible metadata values or `None`, and field hints are normalized into schema-compatible `name`, `comments`, `vendor`, and string fields before being applied. The Syft plugin was updated to emit `NameEntry`, `CommentEntry`, and RFC 3339 timestamps, and the plugin documentation was refreshed to describe the new hook behavior.

* Hardened merge, serialization, and CLI behavior around the new model. Merge now rejects colliding hash data, preserves and merges relationship comments, rewrites and de-duplicates merged `containerPath` values, routes newly added software through the normal index/FS-tree update path, and rebuilds symlink edges from merged metadata after in-place software merges. `cli add` now validates `captureTime` before deserialization, install-path augmentation de-duplicates derived paths, and `stat` now reads through the configured input plugin rather than assuming raw CyTRICS JSON.

* Updated supporting scripts, docs, and test dependencies to match the new workflow. `merge_sbom.py` is now a direct typed CyTRICS merge helper with stdin/stdout support plus root/cycle logging, `merge_additional_metadata.py` has stricter sidecar handling, `merge_config.json` was removed, and the test dependency set now includes `pytest-asyncio` and `jsonschema[format-nongpl]`.

* Refreshed fixtures and tests for 1.0.1 compliance. Sample SBOMs now include the required root fields, remove legacy root and per-software fields, convert `captureTime` values from epoch integers to RFC 3339 strings, and update relationship/unit tests to use valid UUIDs plus hashes or `notHashable` where required. New schema-focused and capture-time-focused tests were added, along with updated CLI, merge, file-type, FS-tree, serialization, and relationship coverage.

## Why

The old CyTRICS path still allowed legacy document shapes, legacy CLI/config flows, and timestamp/hash handling that do not match the v1.0.1 schema. This branch aligns the typed model, generated output shape, fixtures, plugin normalization, CLI entry points, and merge utilities with the v1.0.1 schema while keeping normal runtime construction and serialization permissive. Schema-invalid data is now caught by explicit `validate()` calls and pytest JSON Schema coverage rather than being hidden by serialization.

## Reviewer notes

* This is intentionally a breaking cleanup for the typed CyTRICS model: top-level `systems` handling, related merge/config flows, provenance classes, `recordedInstitution`, and legacy software component wrappers are removed because they are not part of the v1.0.1 schema shape targeted by the typed model, explicit validators, fixtures, and schema tests.

* User-visible command behavior changed with that cleanup. `generate` no longer exposes `--recorded_institution`. `merge_sbom.py` no longer supports `--config_file`, and `merge` no longer supports top-level system creation/wrapping options such as `--system_uuid`, `--system_relationship`, and `--add_system`.

* Plugin authors should review the hook contract changes. `identify_file_type` may now return multiple file-type matches, and `extract_file_info` may return a JSON-compatible metadata value or `None`; plugin docs and built-in file-type ID plugins were updated to reflect that contract.

* Generated `captureTime` values are now RFC 3339 `date-time` strings with timezone information. Explicit validation and schema-focused tests reject legacy epoch-style timestamps.

* Sidecar metadata merging is stricter, and merge semantics are safer: malformed additional metadata now errors, unmatched `sha256hash` values now fail fast, software entries with colliding hash data are refused during merge, and relationship comment data is preserved when duplicate relationships are merged.

## Testing

Updated CLI, merge, file-type, FS-tree, serialization, Java/.NET/ELF/PE relationship, fixture, schema, and capture-time tests to match the v1.0.1 model and serialization rules. The test dependency set was also expanded to include `pytest-asyncio` and `jsonschema[format-nongpl]`.


scripts/README-merge_sbom.md
```diff
@@ -1,10 +1,12 @@
-A folder containing multiple separate SBOM JSON files can be combined using `merge_sbom.py`
-with a command such the one below that gets a list of files using `ls`, and then uses `xargs`
-to pass the resulting list of files to `merge_sbom.py` as arguments.
+Use the main Surfactant `merge` command for the supported merge workflow. The
```
nightlark, line 1: This sounds like good advice. I'm thinking about if we should just remove the merge_sbom.py script in the next release.

scripts/merge_sbom.py
```diff
@@ -4,593 +4,63 @@
 # SPDX-License-Identifier: MIT
```
nightlark, line 4:
```suggestion
# SPDX-License-Identifier: MIT
# /// script
# dependencies = [
#   "surfactant>=0.0.0rc18",
# ]
# ///
```

scripts/merge_additional_metadata.py
```diff
@@ -4,10 +4,13 @@
 # SPDX-License-Identifier: MIT
```
nightlark, line 4:
```suggestion
# SPDX-License-Identifier: MIT
# /// script
# dependencies = [
#   "surfactant>=0.0.0rc18",
# ]
# ///
```

surfactant/cmd/generate.py
```diff
@@ -81,46 +172,64 @@ def get_software_entry(
...
 
-        # ensure metadata exists for the software entry
-        if sw_entry.metadata is None:
-            sw_entry.metadata = []
-        sw_entry.metadata.append(file_details)
+        if not isinstance(file_details, dict):
+            raise TypeError(
+                "extract_file_info plugins must return a metadata object (dict) or None; "
+                f"got {type(file_details).__name__} for {filepath}"
+            )
```
nightlark: This sounds most useful for plugin developers rather than users - would be great if we could create some form of validator that plugin authors could run to check their plugins for issues.

surfactant/cmd/generate.py
```diff
@@ -81,46 +172,64 @@ def get_software_entry(
...
-            for vendor in value:
-                if vendor not in sw_entry.vendor:
-                    sw_entry.vendor.append(vendor)
+            normalized_vendors = _normalize_vendor_hints(value, filepath=filepath)
+            if normalized_vendors:
+                merged_vendors = [*(sw_entry.vendor or [])]
+                for vendor in normalized_vendors:
+                    if vendor not in merged_vendors:
+                        merged_vendors.append(vendor)
+                sw_entry.update_field("vendor", merged_vendors)
```
nightlark, line 313: This would probably also be a good way to handle hints for the new softwareType field, potentially bypassing the confidence field and just adding a check to avoid duplicates.

surfactant/cmd/generate.py
```diff
@@ -225,12 +334,6 @@ def get_default_from_config(option: str, fallback: Optional[Any] = None) -> Any:
     required=False,
     help="Skip including install path information if not given by configuration",
 )
-@click.option(
-    "--recorded_institution",
-    is_flag=False,
-    default=get_default_from_config("recorded_institution"),
-    help="Name of user's institution",
-)
```
nightlark, line 233: We should probably add an option to set the author (name+type).

surfactant/cmd/stat.py
```diff
@@ -1,46 +1,60 @@
...
+
+    num_pe_exe_str = f"Number of PE Executables: {pe_is_exe} with {clr_exe} using .NET/CLR"
+    num_dll_str = f"Number of DLLs: {pe_is_dll} with {clr_dll} using .NET/CLR"
+    num_elf_bin_str = f"Number of ELF Binaries: {elf_is_exe}"
+    num_elf_shared_lib_str = f"Number of ELF shared libraries: {elf_is_lib}"
+
+    click.echo(num_pe_exe_str)
+    click.echo(num_dll_str)
+    click.echo(num_elf_bin_str)
+    click.echo(num_elf_shared_lib_str)
```
nightlark: It would probably be fine to skip the assignment and just echo the format strings directly.

surfactant/plugin/hookspecs.py
```diff
@@ -39,13 +45,18 @@ def extract_file_info(
...
-    anything to the Software entry metadata.
+) -> Optional[Dict[str, Any]]:
+    """Extract information from the given file and add it to the given software entry.
+
+    Plugins may:
+      - return a metadata object to append to ``software.metadata``
+      - add child Software entries through ``children``
+      - provide candidate field values through ``software_field_hints``
+
+    The returned metadata value must be a ``dict`` representing a JSON object, or
```
nightlark: Strictly speaking, the JSON schema allows metadata to be any JSON object, so the value returned here does not need to be a `dict`.

surfactant/sbomtypes/_author.py
```diff
@@ -0,0 +1,18 @@
+# Copyright 2023 Lawrence Livermore National Security, LLC
```
nightlark:
```suggestion
# Copyright 2026 Lawrence Livermore National Security, LLC
```

surfactant/sbomtypes/_comment.py
```diff
@@ -0,0 +1,49 @@
+# Copyright 2023 Lawrence Livermore National Security, LLC
```
nightlark:
```suggestion
# Copyright 2026 Lawrence Livermore National Security, LLC
```

surfactant/cmd/generate.py
```diff
@@ -81,46 +172,64 @@ def get_software_entry(
...
+    if vendor_hints:
+        field_confidence["vendor"] = (vendor_hints, 0)
+
     # set any fields that haven't been set yet (user/previously set fields take precedence)
     for field, (value, _) in field_confidence.items():
         if field == "name" and not sw_entry.name:
-            sw_entry.name = value
+            normalized_names = _normalize_name_hints(value, filepath=filepath)
+            if normalized_names:
+                sw_entry.update_field("name", normalized_names)
```
nightlark: In the new schema, this can be a list of names... we might want to add something to distinguish between different types of names like purl, product name, etc, and pick the highest confidence option of each type?

surfactant/cmd/generate.py
```diff
@@ -81,46 +172,64 @@ def get_software_entry(
...
+                sw_entry.update_field("vendor", merged_vendors)
         elif field == "description" and not sw_entry.description:
-            sw_entry.description = value
+            normalized_description = _normalize_string_hint("description", value, filepath=filepath)
+            sw_entry.update_field("description", normalized_description)
         elif field == "comments" and not sw_entry.comments:
-            sw_entry.comments = value
+            normalized_comments = _normalize_comment_hints(value, filepath=filepath)
+            if normalized_comments:
+                sw_entry.update_field("comments", normalized_comments)
```
nightlark: Comments can probably also be handled similarly to vendors, and skip the confidence level ranking with just a check to avoid duplicates.

surfactant/sbomtypes/_file.py
```diff
@@ -5,15 +5,46 @@
...
 
 @dataclass
 class File:
-    filePath: str
-    description: str
-    category: str
-    capturedBy: str
-    captureTime: str
-    source: str
+    filePath: Optional[str] = None
```
nightlark:
```suggestion
    filePath: str
```

surfactant/sbomtypes/_file.py
```diff
@@ -5,15 +5,46 @@
...
+    capturedBy: str = ""
+    source: str = ""
     methodOfAcquisition: Optional[List[str]] = None
+
+    def __post_init__(self) -> None:
+        """Validate against the CyTRICS file schema requirements."""
+
+        if self.filePath is not None:
+            if not isinstance(self.filePath, str):
+                raise TypeError("filePath must be a string or None")
```
nightlark: According to the schema, `filePath` can't be None.

surfactant/sbomtypes/_hardware.py
```diff
@@ -4,31 +4,114 @@
...
     countryOfOriginSource: Optional[str] = None
     quantity: Optional[int] = None
     description: Optional[str] = None
     vendor: Optional[List[str]] = None
     identifiers: Optional[List[str]] = None
     hardwareType: Optional[List[str]] = None
-    comments: Optional[str] = None
-    metadata: Optional[List[object]] = None
+    comments: Optional[List[CommentEntry]] = None
+    metadata: List[Dict[str, Any]] = field(default_factory=list)
```
nightlark: Items in the metadata list aren't constrained to being a `dict`.

surfactant/sbomtypes/_hardware.py
```diff
@@ -4,31 +4,114 @@
...
+            for item in self.supplementaryFiles:
+                if not isinstance(item, File):
+                    raise TypeError("All items in supplementaryFiles must be File objects")
+
+        if not isinstance(self.metadata, list):
+            raise TypeError("metadata must be a list")
+
+        for item in self.metadata:
+            if not isinstance(item, dict):
+                raise TypeError("All items in metadata must be objects (dicts)")
```
nightlark: This is not true -- they must be a JSON object, which includes more types than just a dict.

surfactant/sbomtypes/_name.py
```diff
@@ -0,0 +1,40 @@
...
+@dataclass
+class NameEntry:
+    nameValue: Optional[str] = None
+    nameType: Optional[str] = None
+
+    def __post_init__(self) -> None:
+        if self.nameValue is not None and not isinstance(self.nameValue, str):
+            raise TypeError("nameValue must be a string or None")
+        if self.nameType is not None and not isinstance(self.nameType, str):
+            raise TypeError("nameType must be a string or None")
```
nightlark: Looking at the schema, nameValue and nameType must be strings, not None. They could be an empty string though.

surfactant/sbomtypes/_software.py
```diff
@@ -7,66 +7,172 @@
...
     md5: Optional[str] = None
-    relationshipAssertion: Optional[str] = (
-        None  # enum: Unknown, Root, Partial, Known; default=Unknown
-    )
-    comments: Optional[str] = None
-    metadata: Optional[List[object]] = None
+    notHashable: Optional[bool] = None
+    relationshipAssertion: Optional["RelationshipAssertion"] = RelationshipAssertion.UNKNOWN
+    comments: Optional[List[CommentEntry]] = None
+    metadata: List[Dict[str, Any]] = field(default_factory=list)
```
nightlark: metadata is not constrained to just containing `dict`.

surfactant/sbomtypes/_software.py
```diff
@@ -7,66 +7,172 @@
...
+        if not all(isinstance(item, entry_type) for item in value):
+            raise TypeError(f"All items in {field_name} must be {entry_type.__name__} objects")
+
+    def _validate_metadata(self) -> None:
+        if not isinstance(self.metadata, list):
+            raise TypeError("metadata must be a list")
+
+        for item in self.metadata:
+            if not isinstance(item, dict):
+                raise TypeError("All items in metadata must be objects (dicts)")
```
nightlark: `dict` is too narrow, metadata list can be "anything" that serializes to a valid JSON object (which is broader than just dict).

surfactant/sbomtypes/_software.py
```diff
@@ -7,66 +7,172 @@
...
+    def _update_field(self, field_name: str, value: Any) -> None:
+        if value in ("", " ", None):
+            return
+
+        original_value = getattr(self, field_name)
+
+        setattr(self, field_name, value)
+
+        try:
+            self.__post_init__()
```
nightlark: This feels like a lot of checks to be performing just to update a single field, when most other fields are not changing. I'd be inclined to say we should not use this method when generating an SBOM.

surfactant/sbomtypes/_software.py
```diff
@@ -138,19 +244,20 @@ def merge(self, sw: Software):
...
+                            if fld.name == "containerPath" and isinstance(new_value, str):
                                 if new_value.startswith(sw.UUID):
                                     new_value = new_value.replace(sw.UUID, self.UUID)
-                            if new_value not in current_arr:
-                                current_arr.append(new_value)
+
+                            if new_value not in merged_arr:
+                                merged_arr.append(new_value)
+
+                        self._update_field(fld.name, merged_arr)
```
nightlark, line 251: I'd also be inclined not to use update_field here due to the expensive validation. Merging can potentially happen multiple times before the final object that will be output in the SBOM will be ready.

surfactant/sbomtypes/_sbom.py
```diff
@@ -6,49 +6,242 @@
...
-from ._system import System
+from ._name import NameEntry
+from ._relationship import Relationship
+from ._software import Software
+from ._tool import Tool
 
 INTERNAL_FIELDS = {"software_lookup_by_sha256"}
+_SPEC_VERSION_RE = re.compile(r"^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)$")
+_CYTRICS_SCHEMA_VERSION = "1.0.1"
+_CYTRICS_SCHEMA_RELATIVE_PATH = Path("docs") / "cytrics_schema" / "schema.json"
```
nightlark: This file will not be accessible to end users.

surfactant/cmd/cli_commands/cli_base.py
```diff
@@ -46,15 +44,7 @@ def serialize(sbom: SBOM) -> str:
...
         """
-        # NOTE: python pickle cannot pickle MappingProxyType, which is inherently included in the Field type.
-        # Pickling is much faster than converting to json or msgpack (see MR for timings: https://github.com/LLNL/Surfactant/pull/261
-        # To workaround the issue we replace the metadata attribute of the Field type (which is default mappingproxytype) with an empty dict
-        # Upon deserialization, we recreate the class with dataclasses.replace() to ensure the unpickled class instance is intact. Without this,
-        # the class function to_json() and to_dict() does not work as the Field type is no longer recongized.
         if isinstance(sbom, SBOM):
-            for _, v in sbom.__dataclass_fields__.items():
-                if isinstance(v, Field):
-                    v.metadata = {}
```
nightlark, line 57: I'm not sure I'm following why this step is no longer necessary. If it isn't directly related to the CyTRICS schema change, I think it would be better to leave it out/do it as a separate PR.

nightlark (changes requested):
The main change I think we should make is move the schema validation to just happening during tests, and perhaps the "field"/post_init validation to an explicit set of "validate" function calls for people using Surfactant as a library.

Reason being that we control the data going into the SBOM we are generating, so it had better pass validation and we don't need to be overly paranoid about checking every single value to make sure its valid -- and if the SBOM we output doesn't pass validation its a bug we (or a plugin author) needs to fix.

Afaik there shouldn't be any "knobs" an end user generating an SBOM can provide that would make Surfactant output an invalid SBOM, with the exception of providing an input SBOM. But if a user provides a hand-crafted SBOM as input that doesn't conform to the schema that's on them and they can expect to get either an error or an SBOM as output that doesn't pass validation either.

Looking at Syft as an example, they perform validation of the SBOMs they generate against the CycloneDX schema when running unit tests (which we should definitely be doing as well), to catch issues with their SBOM generation. But once they are sure they are outputting valid SBOMs, they don't need to check every single SBOM generated by end users.

nightlark:
For `filePath` field in the supplementary files, we should treat the field as required. That should be changed in an upcoming iteration of the schema.

surfactant/sbomtypes/_tool.py
```diff
@@ -0,0 +1,22 @@
+# Copyright 2023 Lawrence Livermore National Security, LLC
```
nightlark:
```suggestion
# Copyright 2026 Lawrence Livermore National Security, LLC
```

surfactant/cmd/generate.py
willis89pr: Added generate options for author name and author type and wired them into the CyTRICS authors field. They can also be set through core.author_name/core.author_type, and the TUI now exposes matching fields. Docs and tests were updated.

surfactant/cmd/stat.py
willis89pr: I removed the temporary string variables and echo the formatted strings directly.

surfactant/sbomtypes/_software.py
willis89pr: Yes, this should be addressed now. `update_field()` no longer triggers `__post_init__()` or explicit validation; it only skips empty values and assigns the field. Merge also uses `_update_field()` directly, so repeated merge updates avoid full-object validation. Schema conformance is now covered by the pytest schema tests instead of implicit runtime validation during generation/merge.

docs/configuration_files.md
```diff
@@ -36,11 +38,12 @@ On Unix-like platforms (including macOS), the XDG directory specification is fol
...
 to `~/.config`. On Windows, the file is stored in the Roaming AppData folder at `%APPDATA%\\surfactant\\config.toml`.
 
-The file itself is a TOML file, and for the previously mentioned example plugin may look something like this:
+The file itself is a TOML file, and for the previously mentioned author settings may look something like this:
 
 ```toml
 [core]
-recorded_institution = "LLNL"
+author_name = "LLNL"
+author_type = "organization"
```
nightlark, line 46: Idea for a future feature: I think author info can be a list, so perhaps we could add an "authors" config option with a list of tuples, so a user could have [["LLNL", "organization"], ["Ryan", "individual"]] to capture both the organization they are part of and the individual(s) at the organization doing the SBOM creation.

surfactant/plugin/hookspecs.py
```diff
@@ -39,13 +45,18 @@ def extract_file_info(
...
     children: List[Software],
-    software_field_hints: List[Tuple[str, object, int]],
+    software_field_hints: List[Tuple[str, Any, int]],
     omit_unrecognized_types: bool,
-) -> object:
-    """Extracts information from the given file to add to the given software entry. Return an
-    object to be included as part of the metadata field, and potentially used as part of
-    selecting default values for other Software entry fields. Returning `None` will not add
-    anything to the Software entry metadata.
+) -> Optional[Any]:
```
nightlark: `Any` might already include `NoneType`, making the `Optional` redundant.

surfactant/sbomtypes/_file.py
```diff
@@ -5,15 +5,48 @@
...
+        """Validate this file entry against the CyTRICS field constraints."""
+
+        if not isinstance(self.filePath, str):
+            raise TypeError("filePath must be a string")
+
+        if not self.filePath:
+            raise ValueError("filePath is required")
+
+        if len(self.filePath) < 5:
+            raise ValueError("filePath must be at least 5 characters long")
```
nightlark, line 33: I wonder if removing this min length is something that should be suggested as a schema change. The logic was probably that the shortest file name is `x.yyy`, but there can be file names with no extension (or a 2 character extension). Though it seems like it would be kind of silly using an extra short and very non-descriptive file.

surfactant/cmd/generate.py
nightlark: I guess for this we would need to add an if branch for the softwareType field.

surfactant/plugin/hookspecs.py
nightlark: I'm thinking `object` could be the better type for the return value, since from a type checking perspective it enforces very strict type checking (unless an isinstance check is performed), whereas `Any` turns off type checking.

surfactant/sbomtypes/_sbom.py
```diff
@@ -1151,60 +1343,35 @@ def create_software(
...
-
-                    # Add the new system node into the graph
-                    if hasattr(self, "graph"):
-                        self.graph.add_node(system.UUID, type="System")
-
-        # 2) Merge software
+        # 1) Merge software
         if sbom_m.software:
             for sw in sbom_m.software:
                 # NOTE: Do we want to pass in teh UUID here? What if we have two different UUIDs for the same file? Should hashes be required?
```
nightlark:
```suggestion
                # NOTE: Do we want to pass in the UUID here? What if we have two different UUIDs for the same file? Should hashes be required?
```

surfactant/sbomtypes/_sbom.py
```diff
@@ -1151,60 +1343,35 @@ def create_software(
...
                 if existing_sw := self._find_software_entry(
                     uuid=sw.UUID, sha256=sw.sha256, md5=sw.md5, sha1=sw.sha1
                 ):
+                    if Software.check_for_hash_collision(existing_sw, sw):
+                        raise ValueError(
+                            "Refusing to merge software entries with colliding hash data: "
+                            f"{existing_sw.UUID} and {sw.UUID}"
+                        )
```
nightlark: I believe part of the point of the merging software entries from two different SBOMs was combining entries for the same file (e.g. hashes match), which the addition of this collision check breaks. As in, the if statement right before this check is literally looking for a software entry with matching hash information.

If hashes matches it is more likely that they refer to the same file rather than the hashes coincidentally matching (particularly if multiple different types of hashes match, and if we add a check for matching sizes we could further increase our confidence in this).

tests/schema/test_cytrics_schema.py
```diff
@@ -0,0 +1,610 @@
...
+Focus areas:
+- Generated timestamps are RFC 3339 date-time strings in UTC
+- Hardware captureTime accepts RFC 3339 strings and null
+- Software captureTime accepts RFC 3339 strings and null
+- File captureTime accepts RFC 3339 strings and rejects null
+- Supplementary filePath is treated as required ahead of the next schema update
+- NameEntry nameValue/nameType values are strings, not null
+- Invalid comment timestamps are caught by schema tests, not runtime constructors
+- Legacy epoch integers are rejected by schema validation
+"""
```
nightlark: It looks like there are tests that validate more than just time-focused things; I don't think the description needs to restrict it to being focused primarily on validating the capture time.

surfactant/plugin/hookspecs.py
```diff
@@ -61,36 +72,35 @@ def extract_file_info(
...
     Returns:
-        object: An object to be added to the metadata field for the software entry. May be `None` to add no metadata.
+        Optional[Any]: A metadata value to append to ``software.metadata``, or ``None`` to add no metadata.
     """
 
 
 @hookspec
 def establish_relationships(
-    sbom: SBOM, software: Software, metadata
+    sbom: SBOM, software: Software, metadata: Any
```
nightlark:
```suggestion
    sbom: SBOM, software: Software, metadata: object
```

surfactant/sbomtypes/_software.py
```diff
@@ -7,66 +7,163 @@
...
     md5: Optional[str] = None
-    relationshipAssertion: Optional[str] = (
-        None  # enum: Unknown, Root, Partial, Known; default=Unknown
-    )
-    comments: Optional[str] = None
-    metadata: Optional[List[object]] = None
+    notHashable: Optional[bool] = None
+    relationshipAssertion: Optional["RelationshipAssertion"] = RelationshipAssertion.UNKNOWN
+    comments: Optional[List[CommentEntry]] = None
+    metadata: List[Any] = field(default_factory=list)
```
nightlark:
```suggestion
    metadata: Optional[List[object]] = field(default_factory=list)
```

surfactant/sbomtypes/_hardware.py
```diff
@@ -4,31 +4,112 @@
...
     countryOfOriginSource: Optional[str] = None
     quantity: Optional[int] = None
     description: Optional[str] = None
     vendor: Optional[List[str]] = None
     identifiers: Optional[List[str]] = None
     hardwareType: Optional[List[str]] = None
-    comments: Optional[str] = None
-    metadata: Optional[List[object]] = None
+    comments: Optional[List[CommentEntry]] = None
+    metadata: List[Any] = field(default_factory=list)
```
nightlark:
```suggestion
    metadata: Optional[List[object]] = field(default_factory=list)
```

surfactant/plugin/hookspecs.py
nightlark:
```suggestion
) -> object:
```

surfactant/plugin/hookspecs.py
```diff
@@ -61,36 +72,35 @@ def extract_file_info(
...
+        software_field_hints (List[Tuple[str, Any, int]]): List of tuples containing the target software field name,
+            a suggested value for that field, and a confidence level. Plugins can add entries to this list to suggest
+            values for software entry fields. The value with the highest confidence for a field will be selected.
         omit_unrecognized_types (bool): Whether files with types that are not recognized by Surfactant should be
             left out of the SBOM. When a plugin is adding additional context entries to the queue, it should typically
             default to propagating this value to the new context entries that it creates.
 
     Returns:
-        object: An object to be added to the metadata field for the software entry. May be `None` to add no metadata.
+        Optional[Any]: A metadata value to append to ``software.metadata``, or ``None`` to add no metadata.
```
nightlark:
```suggestion
        object: A metadata value to append to ``software.metadata``, or ``None`` to add no metadata.
```

surfactant/plugin/hookspecs.py
```diff
@@ -61,36 +72,35 @@ def extract_file_info(
...
+    software/metadata value that has been passed to it.
 
     Returns a list of relationships to be added to the SBOM.
 
     Args:
         sbom (SBOM): The SBOM object that the Software is part of.
-        software (Software): The Software entry that the metadata object is from.
-        metadata: The metadata object to establish relationships based on.
+        software (Software): The Software entry that the metadata value is from.
+        metadata: The metadata value to establish relationships based on.
```
nightlark:
```suggestion
        metadata (object): The metadata value to establish relationships based on.
```

surfactant/sbomtypes/_file.py
nightlark: Issue #618 captures the changes we want to make to the CyTRICS schema

docs/configuration_files.md
nightlark: Issue #619 has been created to track this.

tests/schema/test_cytrics_schema.py
nightlark:
```suggestion
"""Schema-focused tests for SBOMs using the CyTRICS format.

These tests ensure that generated SBOMs and data validation functions remain aligned
with the checked-in CyTRICS schema at docs/cytrics_schema/schema.json.
"""
```

surfactant/sbomtypes/_sbom.py
willis89pr: I removed the hash-collision rejection to keep the existing merge behavior intact.

I kept the add_software() and refresh changes because they preserve derived SBOM state when merge mutates an existing software entry or inserts a new one. The old code appended software directly and manually added graph nodes, which bypassed software_lookup_by_sha256 and fs_tree updates. Using add_software() for new entries and refreshing merged entries keeps find_software(), install-path lookup, hash edges, and symlink metadata reconstruction consistent with the merged software list.

I also scoped the symlink metadata rebuild so it only runs when software was actually merged or added.

nightlark (approved):

PR #622, opened 2026-05-18
Ruff Pyupgrade, Modernization and other linter rules 

Nova38:
### Summary

If merged this pull request will modernize the codebase using ruff by enabling several of its linters. The fixes were mostly done with the automated fixes. Several of the PTH (use pathlib) required manualy fixing them. Currently all pytest passed locally

### Proposed changes

```toml
select = [
    "B",
    "C4",     # flake8-comprehensions (C4) (https://docs.astral.sh/ruff/rules/#flake8-comprehensions-c4)
    "E",
    "F",
    "FURB",   # refurb (https://docs.astral.sh/ruff/rules/#refurb-furb)
    "I",
    "PIE",    # flake8-pie (PIE) https://docs.astral.sh/ruff/rules/#flake8-pie-pie
    "PTH",    # flake8-use-pathlib (PTH)[https://docs.astral.sh/ruff/rules/#flake8-use-pathlib-pth]
    "PYI",    # flake8-PYI https://docs.astral.sh/ruff/rules/#flake8-pyi-pyi
    "Q",      # flake8-quotes (Q) https://docs.astral.sh/ruff/rules/#flake8-quotes
    "RET",    # flake8-return (RET) (https://docs.astral.sh/ruff/rules/#flake8-raise-rse)
    "RUF010", #
    "RUF013", # PEP 484 prohibits implicit `Optional`
    "RUF019", # [*] Unnecessary key check before dictionary access",
    "RUF022", #`__all__` is not sorted
    "TC",     # flake8-type-checking (TC) (https://docs.astral.sh/ruff/rules/#flake8-type-checking-tc)
    "UP",     # pyupgrade (https://docs.astral.sh/ruff/rules/#pyupgrade-up)
    "W",
]
```


surfactant/infoextractors/rpm_file.py
```diff
@@ -191,10 +191,7 @@ def extract_rpm_info(filename: str) -> Dict[str, Any]:
             file_details["rpm"]["buildtime"] = header["buildtime"]
         if "basenames" in header:
             file_hash_location = ""
-            if "filedigests" in header:
-                file_hash_location = "filedigests"
-            else:
-                file_hash_location = "filemd5s"
+            file_hash_location = "filedigests" if "filedigests" in header else "filemd5s"
```
matthewkelley22, line 193: While I understand why the modification was made, I feel like this modification makes things less readable.

surfactant/cmd/cli_commands/__init__.py
```diff
@@ -7,4 +7,4 @@
 from .cli_load import Load
 from .cli_save import Save
 
-__all__ = ["Load", "Save", "Cli"]
+__all__ = ["Cli", "Load", "Save"]
```
matthewkelley22, line 10: Pretty sure that the reordering won't hurt anything, but I'm not familiar with this part of the codebase.

.pre-commit-config.yaml
matthewkelley22, line 1: Trusting based on CI

surfactant/fileinfo.py
```diff
@@ -18,7 +18,7 @@ def get_file_info(filename):
         Optional[dict]: Dictionary that contains info about the file.
     """
     try:
-        fstats = os.stat(filename)
+        fstats = os.stat(filename)  # noqa: PTH116
```
matthewkelley22: Why are we using the `os` version of `stat` instead of the `pathlib` version?

surfactant/plugin/manager.py
```diff
@@ -111,9 +111,7 @@ def is_plugin_blocked(pm: pluggy.PluginManager, plugin_name: str) -> bool:
     Returns:
         bool: True if the plugin is blocked, False otherwise.
     """
-    if pm.is_blocked(plugin_name):
-        return True
-    return False
+    return bool(pm.is_blocked(plugin_name))
```
matthewkelley22:
```suggestion
    return pm.is_blocked(plugin_name)
```

matthewkelley22: `is_blocked()` already returns a `bool`

surfactant/plugin/hookspecs.py
```diff
@@ -3,21 +3,21 @@
...
-from queue import Queue
-from typing import Any, List, Optional, Tuple, Union
+from typing import TYPE_CHECKING, Any
 
 from pluggy import HookspecMarker
 
 from surfactant import ContextEntry
 from surfactant.sbomtypes import SBOM, Relationship, Software
 
+if TYPE_CHECKING:
```
matthewkelley22, line 13: Pretty sure we need `Queue` at runtime. Not sure why this didn't cause any issues in the test cases tbh. Maybe this is true when running tests?

surfactant/filetypeid/id_magic.py
```diff
@@ -255,4 +251,4 @@ def identify_file_type(
     except (FileNotFoundError, PermissionError):
         return None
 
-    return filetype_matches if filetype_matches else None
```
matthewkelley22, line 258: Not super familiar with this syntax, trusting based on CI

matthewkelley22 (commented):
Have some feedback for review, didn't look at the larger changes or anything involving regex/something I wasn't super familiar with.

surfactant/cmd/cli_commands/__init__.py
nightlark: Ordering doesn't matter here, `__all__` is just listing what should be imported when a wildcard is used (`from blah import *`).

surfactant/infoextractors/rpm_file.py
```diff
@@ -191,10 +191,7 @@ def extract_rpm_info(filename: str) -> Dict[str, Any]:
             file_details["rpm"]["buildtime"] = header["buildtime"]
         if "basenames" in header:
             file_hash_location = ""
```
nightlark:
```suggestion
```

surfactant/infoextractors/rpm_file.py
nightlark: Mostly stylistic, I think we can just go with what ruff put since it's a simple enough condition, and the other surrounding code isn't particularly dense. (Also avoids fighting with ruff trying to change it back again if we do another pyupgrade in the future).

surfactant/plugin/hookspecs.py
nightlark: It's not an error because hookspecs.py and plugins implementing hooks that provide a queue as one of the arguments are not actually running any code from the queue module -- it's just being used as a type hint to communicate to users (and static type checking tools) what the type should be. The hint is just metadata and never checked at runtime by the Python interpreter.

It might even better to turn the type hint that the Queue is being used into a string by wrapping it in quotes. Pluggy also adds a layer of indirection that kinda messes with static type checking.

The only place the import queue is needed is I think in generate.py where the hook gets called from (since that code is the one that needs to create a new instance of a queue).

surfactant/filetypeid/id_magic.py
nightlark: The `or` operator returns the first "Truthy" value it finds, or the last one if none of the preceding ones are truthy. (Imagine `or` as being like a special way of calling a function where the arguments on either side of the function name rather than a comma-separated list inside parenthesis like `func(arg1, arg2)`.)

Instead of being in a return statement, we could do something like `myvar = filetype_matches or None` to initialize `myvar` to filetype_matches if it is true, and otherwise set it to None rather than false. But its more flexible than that, since Truthy values in Python include things like non-empty lists. So something like this could set `myvar` to the list if it contains stuff, and otherwise make it `None` rather than an empty list:

```python
mylist = ["abc"]
myvar = mylist or None
```

docs/conf.py
```diff
@@ -66,30 +66,32 @@
...
-                with open(os.path.join(image_dir, file_name), "wb") as img_file:
+                with (image_dir / file_name).open("wb") as img_file:
                     img_file.write(response.content)
             else:
                 print(f"Failed to download {url}")
 
 
 # Path to the TOML file
-toml_file_path = os.path.join(os.path.dirname(__file__), "images.toml")
+toml_file_path = Path(__file__).parent / "images.toml"
```
nightlark:
```suggestion
toml_file_path = base_dir / "images.toml"
```

docs/conf.py
```diff
@@ -66,30 +66,32 @@
...
 # Directory to save the images
-image_directory = os.path.join(os.path.dirname(__file__), "img")
+image_directory = Path(__file__).parent / "img"
```
nightlark:
```suggestion
image_directory = base_dir / "img"
```

docs/conf.py
```diff
@@ -106,24 +108,27 @@ def download_images_from_toml(toml_file, image_dir):
 # Make CyTRICS schema available as a static file under cytrics_schema/
 # -------------------------------------------------------------------
 def _copy_cytrics_schema(app, exception):
-    src = os.path.abspath(os.path.join(os.path.dirname(__file__), "cytrics_schema", "schema.json"))
-    if not os.path.exists(src):
+    src = Path(__file__).parent / "cytrics_schema" / "schema.json"
```
nightlark:
```suggestion
    src = base_dir / "cytrics_schema" / "schema.json"
```

docs/conf.py
```diff
@@ -106,24 +108,27 @@ def download_images_from_toml(toml_file, image_dir):
...
 
 # -------------------------------------------------------------------
 # Make capa rules available in the capa/ subfolder
 # -------------------------------------------------------------------
 def _setup_capa_directory(app, exception):
     # Copy local capa files to output directory
-    src_dir = os.path.join(os.path.dirname(__file__), "capa")
-    dst_dir = os.path.join(app.outdir, "capa")
-    if os.path.exists(src_dir):
+    src_dir = Path(__file__).parent / "capa"
```
nightlark:
```suggestion
    src_dir = base_dir / "capa"
```

surfactant/infoextractors/ole_file.py
```diff
@@ -150,7 +153,7 @@ def extract_msi(filename: str, output_folder: str) -> List[Tuple[str, str]]:
 
 
 def preprocess_msi_decompression(msi: pymsi.Msi):
-    folders: List[CabFolder] = []
+    folders: list[CabFolder] = []
```
nightlark:
```suggestion
    folders: "list[CabFolder]" = []
```

surfactant/output/csv_writer.py
```diff
@@ -78,10 +77,9 @@ def get_software_field(software, field):
...
         return getattr(software, field)
     # Copyright field currently only gets populated from Windows PE file metadata
-    if field == "Copyright":
-        if software.metadata and isinstance(software.metadata, Iterable):
-            retval = []
-            for entry in software.metadata:
-                if "FileInfo" in entry and "LegalCopyright" in entry["FileInfo"]:
-                    return entry["FileInfo"]["LegalCopyright"]
+    if field == "Copyright" and software.metadata and isinstance(software.metadata, Iterable):
+        retval = []
```
nightlark:
```suggestion
```

surfactant/utils/exit_hook.py
```diff
@@ -1,13 +1,16 @@
...
 
 # https://stackoverflow.com/a/9741784
 class ExitHooks:
     def __init__(self):
-        self.exit_code: Optional[int] = None
-        self.exception: Optional[Exception] = None
-        self._orig_exit: Optional[Callable[[int], None]] = None
+        self.exit_code: int | None = None
+        self.exception: Exception | None = None
+        self._orig_exit: Callable[[int], None] | None = None
```
nightlark:
```suggestion
        self._orig_exit: "Callable[[int], None] | None" = None
```

surfactant/utils/regex.py
```diff
@@ -574,11 +574,8 @@ def extract_internal_literals(node, max_possibilities, min_length, parser):
         # Only consider required repetitions
         if min_count > 0:
             sub_tree = node.children[0]
-            sub_literals = extract_internal_literals(
-                sub_tree, max_possibilities, min_length, parser
-            )
+            return extract_internal_literals(sub_tree, max_possibilities, min_length, parser)
 
             # If always repeated at least once, these literals must be present
```
nightlark:
```suggestion
            # If always repeated at least once, these "sub"-literals must be present
            return extract_internal_literals(sub_tree, max_possibilities, min_length, parser)
```

surfactant/fileinfo.py
nightlark:
```suggestion
        fstats = Path(filename).stat()
```

nightlark (approved):

PR #642, opened 2026-06-22
Generate NuGet PURL's

KendallHarterAtWork:
### Summary

If merged this pull request will generate PURL's for NuGet packages as an additional name in the SBOM.

### Proposed changes

This also runs `init_hook` for hooks implementing `establish_relationships` during the generate command, which was not done previously.

### DIFF
--- a/surfactant/relationships/nuget_purl.py
+++ b/surfactant/relationships/nuget_purl.py
@@ -68,7 +68,7 @@ class _NuGetManager:
                 r = requests.get(
                     f"{url}/{pn_low}/{ver_low}/{pn_low}.{ver_low}.nupkg",
                     stream=True,
-                    timeout=self.request_timeout
+                    timeout=self.request_timeout,
                 )
             except requests.RequestException as e:
                 logger.warning(f"NuGet download failed for {pn_low}.{ver_low}.nupkg - {e}")
@@ -153,9 +153,7 @@ def init_hook(command_name: str | None = None):
         return
 
     if not ConfigManager().get("nuget", "enable_lookups", False):
-        logger.info(
-            "[nuget_purl] NuGet lookups disabled via config (nuget.enable_lookups=false)"
-        )
+        logger.info("[nuget_purl] NuGet lookups disabled via config (nuget.enable_lookups=false)")
         _nuget.disabled = True
         return
 
@@ -182,6 +180,7 @@ def establish_relationships(sbom: SBOM, software: Software, metadata) -> list[Re
                     if software.name is None:
                         software.name = []
                     software.name.append(NameEntry(purl, "PURL"))
+
 
 @surfactant.plugin.hookimpl
 def settings_name() -> str | None:
--- a/surfactant/relationships/nuget_purl.py
+++ b/surfactant/relationships/nuget_purl.py
@@ -3,6 +3,14 @@
 #
 # SPDX-License-Identifier: MIT
 
+"""
+Config Options:
+    enable_lookups(bool):
+        Enable NuGet network requests, default is False.
+    request_timeout(int):
+        Time to wait for a NuGet network response in seconds, default is 30.
+"""
+
 import io
 import pathlib
 import zipfile
@@ -11,25 +19,41 @@ import requests
 from loguru import logger
 
 import surfactant.plugin
+from surfactant.configmanager import ConfigManager
 from surfactant.sbomtypes import SBOM, NameEntry, Relationship, Software
 
 
-class __NuGetManager:
+class _NuGetManager:
     def __init__(self):
         self.disabled = True
         self.package_base_addresses = []
 
+        config_manager = ConfigManager()
+        self.request_timeout = float(config_manager.get("nuget", "request_timeout", 30.0))
+
+        # package_name.lower() -> version list, or None if the package wasn't
+        # found. None is cached to avoid re-querying misses.
+        self._index_cache: dict[str, list[str] | None] = {}
+
+        # (package_name.lower(), version.lower()) -> set of nupkg member basenames.
+        # An empty set means download failed/not found; cached so the same
+        # nupkg is never re-downloaded during an enrichment pass.
+        self._members_cache: dict[tuple[str, str], set[str]] = {}
+
     def init_urls(self):
         # Get the base PackageBaseAddress URL
-        r = requests.get("https://api.nuget.org/v3/index.json")
-        if r.status_code != 200:
-            logger.warning(f"NuGet API returned {r.status_code}; disabling")
+        try:
+            r = requests.get("https://api.nuget.org/v3/index.json", timeout=self.request_timeout)
+            r.raise_for_status()
+            resources = r.json()["resources"]
+        except (requests.RequestException, ValueError, KeyError) as e:
+            logger.warning(f"NuGet API unavailable ({e}); disabling NuGet lookups")
             self.disabled = True
             return
 
         self.disabled = False
         self.package_base_addresses = [
-            x["@id"] for x in r.json()["resources"] if x["@type"] == "PackageBaseAddress/3.0.0"
+            x["@id"] for x in resources if x["@type"] == "PackageBaseAddress/3.0.0"
         ]
         # remove trailing "/" if present
         for i, pba in enumerate(self.package_base_addresses):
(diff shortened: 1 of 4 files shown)


surfactant/relationships/nuget_purl.py
```diff
@@ -0,0 +1,114 @@
...
+    """Checks NuGet for a package name and adds it as a name if it exists"""
+
+    if __nuget.disabled:
+        return
+
+    if "dotnetAssembly" not in metadata:
+        logger.debug(
+            f"[nuget_purl] Skipping: No dotnetAssembly info for NuGet PURL in {software.UUID}"
+        )
+        return
```
nightlark:
```suggestion
    if not isinstance(metadata, dict) or "dotnetAssembly" not in metadata:
        logger.debug(
            f"[nuget_purl] Skipping: No dotnetAssembly info for NuGet PURL in {software.UUID}"
        )
        return None
```

Check to make sure metadata entry is a dict before looking for a key in it.

surfactant/relationships/nuget_purl.py
```diff
@@ -0,0 +1,114 @@
...
+import zipfile
+
+import requests
+from loguru import logger
+
+import surfactant.plugin
+from surfactant.sbomtypes import SBOM, NameEntry, Relationship, Software
+
+
+class __NuGetManager:
```
nightlark:
```suggestion
class _NuGetManager:
```

Single underscore is fine/consistent with other parts of the code.

surfactant/relationships/nuget_purl.py
```diff
@@ -0,0 +1,114 @@
...
+                else:
+                    # Unknown package version; check the latest package version
+                    latest_version = versions[-1]
+                    if self.file_is_in_package(file_name, package_name, latest_version):
+                        return f"pkg:nuget/{package_name}"
+
+        return None
+
+
+__nuget = __NuGetManager()
```
nightlark:
```suggestion
_nuget = _NuGetManager()
```

surfactant/relationships/nuget_purl.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+@surfactant.plugin.hookimpl
+def init_hook(command_name: str | None = None):
+    __nuget.init_urls()
```
nightlark:
```suggestion
    if command_name != "generate":
        return

    # Give users an option to disable network lookups
    if not ConfigManager().get("nuget", "enable_lookups", False):
        logger.info(
            "[nuget_purl] NuGet lookups disabled via config (nuget.enable_lookups=false)"
        )
        _nuget.disabled = True
        return
    _nuget.init_urls()
```

Gate the initialization on the "generate" command.

Adding an option for users to control network lookups (outside of installing/uninstalling the plugin -- could potentially leverage info from dapper plugin as an offline way to add some PURLs). Making this opt-in by default to avoid surprise outgoing network requests.

surfactant/relationships/nuget_purl.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+@surfactant.plugin.hookimpl
+def establish_relationships(sbom: SBOM, software: Software, metadata) -> list[Relationship] | None:
+    """Checks NuGet for a package name and adds it as a name if it exists"""
+
+    if __nuget.disabled:
```
nightlark:
```suggestion
    if _nuget.disabled:
```

surfactant/relationships/nuget_purl.py
```diff
@@ -0,0 +1,114 @@
...
+    if "dotnetAssembly" not in metadata:
+        logger.debug(
+            f"[nuget_purl] Skipping: No dotnetAssembly info for NuGet PURL in {software.UUID}"
+        )
+        return
+
+    for dna in metadata["dotnetAssembly"]:
+        if software.fileName:
+            for name in software.fileName:
+                if purl := __nuget.get_package_url(name, dna["Name"], dna["Version"]):
```
nightlark:
```suggestion
                if purl := _nuget.get_package_url(name, dna["Name"], dna["Version"]):
```

surfactant/relationships/nuget_purl.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+class __NuGetManager:
+    def __init__(self):
+        self.disabled = True
+        self.package_base_addresses = []
+
+    def init_urls(self):
+        # Get the base PackageBaseAddress URL
+        r = requests.get("https://api.nuget.org/v3/index.json")
```
nightlark:
```suggestion
        r = requests.get("https://api.nuget.org/v3/index.json", timeout=_REQUEST_TIMEOUT)
```

surfactant/relationships/nuget_purl.py
```diff
@@ -0,0 +1,114 @@
...
+import pathlib
+import zipfile
+
+import requests
+from loguru import logger
+
+import surfactant.plugin
+from surfactant.sbomtypes import SBOM, NameEntry, Relationship, Software
+
+
```
nightlark, line 27:
```suggestion
#  Network calls are best-effort; how long to wait to avoid hanging.
_REQUEST_TIMEOUT = 30

```

surfactant/relationships/nuget_purl.py
```diff
@@ -0,0 +1,114 @@
...
+import io
+import pathlib
+import zipfile
+
+import requests
+from loguru import logger
+
+import surfactant.plugin
+from surfactant.sbomtypes import SBOM, NameEntry, Relationship, Software
```
nightlark, line 25:
```suggestion
import surfactant.plugin
from surfactant.configmanager import ConfigManager
from surfactant.sbomtypes import SBOM, NameEntry, Relationship, Software
```

Import config manager to use it to give user an option to specifically disable NuGet network requests

surfactant/relationships/nuget_purl.py
```diff
@@ -0,0 +1,114 @@
...
+        return None
+
+
+__nuget = __NuGetManager()
+
+
+@surfactant.plugin.hookimpl
+def init_hook(command_name: str | None = None):
+    __nuget.init_urls()
+
```
nightlark, line 171:
```suggestion


@surfactant.plugin.hookimpl
def settings_name() -> str | None:
    return "nuget"
```

surfactant/relationships/nuget_purl.py
```diff
@@ -0,0 +1,114 @@
...
+        r = requests.get("https://api.nuget.org/v3/index.json")
+        if r.status_code != 200:
+            logger.warning(f"NuGet API returned {r.status_code}; disabling")
+            self.disabled = True
+            return
+
+        self.disabled = False
+        self.package_base_addresses = [
+            x["@id"] for x in r.json()["resources"] if x["@type"] == "PackageBaseAddress/3.0.0"
+        ]
```
nightlark, line 59:
```suggestion
        # Get the base PackageBaseAddress URL
        try:
            r = requests.get("https://api.nuget.org/v3/index.json", timeout=_REQUEST_TIMEOUT)
            r.raise_for_status()
            resources = r.json()["resources"]
        except (requests.RequestException, ValueError, KeyError) as e:
            logger.warning(f"NuGet API unavailable ({e}); disabling NuGet lookups")
            self.disabled = True
            return

        self.disabled = False
        self.package_base_addresses = [
            x["@id"] for x in resources if x["@type"] == "PackageBaseAddress/3.0.0"
        ]
```

Catch other potential network related exceptions, and the off chance that something strange happens and valid JSON is not returned.

surfactant/relationships/nuget_purl.py
```diff
@@ -0,0 +1,114 @@
...
+
+import surfactant.plugin
+from surfactant.sbomtypes import SBOM, NameEntry, Relationship, Software
+
+
+class __NuGetManager:
+    def __init__(self):
+        self.disabled = True
+        self.package_base_addresses = []
+
```
nightlark, line 32:
```suggestion
        # package_name.lower() -> version list, or None if the package wasn't
        # found. None is cached to avoid re-querying misses.
        self._index_cache: dict[str, list[str] | None] = {}
        
        # (package_name.lower(), version.lower()) -> set of nupkg member basenames.
        # An empty set means download failed/not found; cached so the same
        # nupkg is never re-downloaded during an enrichment pass.
        self._members_cache: dict[tuple[str, str], set[str]] = {}
```

Add caching to avoid re-downloading the same data multiple times in an enrichment pass.

surfactant/relationships/nuget_purl.py
```diff
@@ -0,0 +1,114 @@
...
+        return None
+
+    def file_is_in_package(self, file_name: str, package_name: str, package_version: str) -> bool:
+        if nuget := self.download_nuget(package_name, package_version):
+            for f in nuget.infolist():
+                if pathlib.Path(f.filename).name == file_name:
+                    return True
+        return False
+
+    def get_package_url(
```
nightlark, line 125:
```suggestion
    def _get_versions(self, package_name: str) -> list[str] | None:
        """Return a package's version list, caching the result.

        Queries each PackageBaseAddress until one serves the package's
        index.json. The result — including a negative None — is cached per
        package name, so repeated files from the same package cost one request.
        """
        key = package_name.lower()
        if key in self._index_cache:
            return self._index_cache[key]

        versions: list[str] | None = None
        for url in self.package_base_addresses:
            try:
                r = self.session.get(f"{url}/{key}/index.json", timeout=_REQUEST_TIMEOUT)
                if r.status_code != 200:
                    continue
                versions = r.json().get("versions") or None
            except (requests.RequestException, ValueError) as e:
                logger.warning(f"NuGet index lookup failed for {key} - {e}")
                continue
            if versions:
                break

        self._index_cache[key] = versions
        return versions

    def get_package_url(
```

Add a helper for getting package versions that caches response.

surfactant/relationships/nuget_purl.py
```diff
@@ -0,0 +1,114 @@
...
+                logger.warning(f"Could not unpack {pn_low}.{ver_low}.nupkg - {e}")
+        return None
+
+    def file_is_in_package(self, file_name: str, package_name: str, package_version: str) -> bool:
+        if nuget := self.download_nuget(package_name, package_version):
+            for f in nuget.infolist():
+                if pathlib.Path(f.filename).name == file_name:
+                    return True
+        return False
+
```
nightlark:
```suggestion
        key = (package_name.lower(), package_version.lower())
        if key not in self._members_cache:
            members: set[str] = set()
            if nuget := self.download_nuget(package_name, package_version):
                members = {pathlib.Path(f.filename).name.lower() for f in nuget.infolist()}
            # Cache even an empty set so a failed/missing download isn't retried.
            self._members_cache[key] = members
        return file_name.lower() in self._members_cache[key]

```

Cache results from looking in zip file to avoid re-downloads when encountering the same package again.

surfactant/relationships/nuget_purl.py
```diff
@@ -0,0 +1,114 @@
...
+            if r.status_code != 200:
+                continue
+            try:
+                # For some reason, have to wrap r.raw (a file-like object)
+                # into an io.BytesIO object to get it to read correctly.
+                # No idea why.
+                return zipfile.ZipFile(io.BytesIO(r.raw.read()))
+            except zipfile.BadZipFile as e:
+                logger.warning(f"Could not unpack {pn_low}.{ver_low}.nupkg - {e}")
+        return None
```
nightlark, line 84:
```suggestion
        pn_low = package_name.lower()
        ver_low = package_version.lower()
        for url in self.package_base_addresses:
            try:
                r = requests.get(
                    f"{url}/{pn_low}/{ver_low}/{pn_low}.{ver_low}.nupkg",
                    stream=True,
                    timeout=_REQUEST_TIMEOUT,
                )
            except requests.RequestException as e:
                logger.warning(f"NuGet download failed for {pn_low}.{ver_low}.nupkg - {e}")
                continue
            if r.status_code != 200:
                continue
            try:
                return zipfile.ZipFile(io.BytesIO(r.raw.read()))
            except zipfile.BadZipFile as e:
                logger.warning(f"Could not unpack {pn_low}.{ver_low}.nupkg - {e}")
        return None
```

Catch request exceptions.

surfactant/relationships/nuget_purl.py
```diff
@@ -0,0 +1,114 @@
...
+                    # Found a matching package version, check that specific version
+                    if self.file_is_in_package(file_name, package_name, package_version):
+                        return f"pkg:nuget/{package_name}@{package_version}"
+                else:
+                    # Unknown package version; check the latest package version
+                    latest_version = versions[-1]
+                    if self.file_is_in_package(file_name, package_name, latest_version):
+                        return f"pkg:nuget/{package_name}"
+
+        return None
```
nightlark:
```suggestion
        versions = self._get_versions(package_name)
        if not versions:
            return None

        if package_version in versions:
            # Found a matching package version, check that specific version
            if self.file_is_in_package(file_name, package_name, package_version):
                return f"pkg:nuget/{package_name}@{package_version}"
        else:
            # Unknown package version; check the latest stable package version if available
            stable = [v for v in versions if "-" not in v] or versions
            latest_version = stable[-1]
            if self.file_is_in_package(file_name, package_name, latest_version):
                return f"pkg:nuget/{package_name}"

        return None
```

Use _get_versions helper, favor latest stable version over pre-releases (unless the package only has pre-release versions).

nightlark (commented):
* Some extra guards/checks against potential exceptions, stylistic single/double underscore naming change.
* Add a user config option to opt-in to network requests to nuget.
* Add caching for results from NuGet API requests to eliminate repeated requests for the same info, and a network timeout to avoid hanging if the server is unresponsive.

I'm thinking in the future maybe this could also leverage any results from dapper that might be in the metadata, which could allow for an "offline" fallback, or reduce the number of NuGet API requests that are needed.

For a future change outside of this PR, we might want to add a new hook type that this could later be updated to use along the lines of
`enrich_software(sbom: SBOM, software: Software) -> None` that would be called once per software entry (perhaps after establishing relationships) rather than once for each metadata entry, and would be intended for plugins that are looking at potentially many metadata entries and relationships at the same time.

nightlark:
Fixed a small regression that prevented the version check from working, and noticed that the metadata["FileInfo"]["ProductVersion"] is often better for lookups than the assembly version (or any of the other version fields in file info.

Quite a few frequently used packages will add fourth version number components that make the NuGet lookup not find a matching version. "ProductVersion" on the other hand appears to always match the version numbers that appear on NuGet including pre-release components (after removing any trailing `+<hash>` from it). I suspect "ProductName" might be the best option for looking up a matching package (possibly even in cases where the file name != package name), if we had some way to map it to the package name to search for on NuGet.

nightlark (approved):

PR #643, opened 2026-06-26
purl builder added to workbench

theStache:
### Summary
If merged this pull request will add a toggle which will display a package url builder that builds a new entry and relationship and adds it to the SBOM.

### Proposed changes
Add purl builder


nightlark:
It looks like there are some more recent changes to the workbench in main that are being reverted by this PR.

I think maybe a diff/patch for the changes between the file with the purl builder and the initial version of sbom-workbench that was passed around to rebase it on the version in git.

PR #646, opened 2026-07-02
Ubuntu rootfs Downloader

KendallHarterAtWork:
### Summary

If merged this pull request will add a plugin that automatically downloads and extracts rootfs files from Ubuntu for ELF files based on the architecture.


plugins/rootfsdownloader/pyproject.toml
```diff
@@ -0,0 +1,34 @@
+[build-system]
+requires = ["setuptools>=77.0.3", "setuptools-scm"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "surfactantplugin-rootfsdownloader"
```
nightlark: Let's just call the plugin `rootfsmanager` (+ rename folder/files/README description) to match the class name used. Thinking about how this could be used, I'd also be inclined to remove the `surfactantplugin-` prefix from the name; I could imagine there being some interest in using it outside of surfactant.

nightlark:
Some notes on other places we can acquire a rootfs from:

* Alpine (musl) "equivalent" of Ubuntu oci: `https://dl-cdn.alpinelinux.org/alpine/` has minirootfs tar files under paths like `https://dl-cdn.alpinelinux.org/alpine/v3.<x>/releases/<arch>/alpine-minirootfs-<ver>-<arch>.tar.gz`
* Void Linux publishes rootfs images for musl and glibc at `https://repo-default.voidlinux.org/live/current/`, unlike alpine which uses busybox, void linux is maybe more similar in their userspace binaries to a desktop distro like debian
* `https://images.linuxcontainers.org` has rootfs images for a variety of distros (not particularly minimal like some of the ones aimed at using for OCI container images)
* `https://github.com/debuerreotype/docker-debian-artifacts` has some rootfs things from Debian
* Arch Linux publishes a bootstrap tarball and has instructions for creating a chroot from it `https://wiki.archlinux.org/title/Install_Arch_Linux_from_existing_Linux`
* Dumping from any Docker container: `podman create --name tmp alpine:3.21 && podman export tmp | gzip > alpine-rootfs.tar.gz`
* Docker images for openwrt rootfs? `https://github.com/openwrt/docker-openwrt/pkgs/container/rootfs` (sorta getting into embedded Linux filesystems... reminds me that eventually support should be added to recognize what OS a filesystem dump is using e.g. /etc/os-release files and package manager metadata files)

plugins/rootfsmanager/pyproject.toml
```diff
@@ -0,0 +1,34 @@
+[build-system]
+requires = ["setuptools>=77.0.3", "setuptools-scm"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "surfactantplugin-rootfsmanager"
```
nightlark:
```suggestion
name = "rootfsmanager"
```
(This is what will determine the name if we push it to pypi).

